PR #11, opened 2025-02-21
feat: switch to openassistant/core for AI module

lixun910:
WIP:
- try openassistant/core to power AI assistant in SQLRooms 
- use local context to avoid sending query results back to LLMs
- use streaming response
- add more provider and models

sqlroom-1


examples/sqlrooms-ai/src/store.ts
```diff
@@ -120,6 +121,14 @@ export const {projectStore, useProjectStore} = createProjectStore<
             structuredOutputs: true,
           });
         },
+        getApiKey: () => {
+          return get()?.openAiApiKey || '';
+        },
+        getTableSchema: async () => {
```
ibgreen: getTableSchemas?

Could this be part of a `project` or `duckdb` slice?

packages/ai/src/AiSlice.ts
```diff
@@ -53,21 +53,25 @@ export type AiSliceState = {
     messagesById: Map<string, AiMessage>;
     addMessages: (messages: AiMessage[]) => void;
     getMessages: () => AiMessage[];
+    setAiModel: (model: string) => void;
+    tableSchema: string;
```
ibgreen:
```suggestion
    tableSchemas: string;
```

packages/ai/src/AiSlice.ts
```diff
@@ -53,21 +53,25 @@ export type AiSliceState = {
     messagesById: Map<string, AiMessage>;
     addMessages: (messages: AiMessage[]) => void;
     getMessages: () => AiMessage[];
+    setAiModel: (model: string) => void;
```
ibgreen, line 59: Nit: Can we add TSDoc over new fields?
```suggestion
    /** Specify AI model. Model names are matched with... */
    setAiModel: (model: string) => void;
```

ibgreen (approved):
LGTM

ibgreen:
@kargeor

examples/sqlrooms-ai/src/store.ts
ilyabo: There's already `getDuckTableSchema` in `@sqlrooms/duckdb`:
https://github.com/sqlrooms/sqlrooms/blob/main/packages/duckdb/src/useDuckDb.ts#L202 
Maybe we can reuse it

examples/sqlrooms-ai/tsconfig.app.json
```diff
@@ -6,6 +6,7 @@
     "lib": ["ES2020", "DOM", "DOM.Iterable"],
     "module": "ESNext",
     "skipLibCheck": true,
+    "sourceMap": true,
```
ilyabo: Thanks!

ilyabo: I also [added it to the base tsconfig](https://github.com/sqlrooms/sqlrooms/commit/549110f9362cd587886abd063c977aee6475f536)

packages/ai/src/AnalysisResult.tsx
```diff
@@ -1,14 +1,15 @@
...
 import {SquareTerminalIcon, CodeIcon} from 'lucide-react';
 import {Button, Popover, PopoverContent, PopoverTrigger} from '@sqlrooms/ui';
+import Markdown from 'react-markdown';
 
 interface AnalysisResultProps {
   result: AnalysisResultSchema;
 }
 
 export const AnalysisResult: React.FC<AnalysisResultProps> = ({result}) => {
+  console.log('result', result);
```
ilyabo: should probably remove debug output

packages/ai/src/analysis.ts
```diff
@@ -15,79 +28,84 @@ import {
...
-    abortSignal,
-    // prompt,
-    messages,
-
-    tools: TOOLS,
-
-    toolChoice: 'required',
+  const assistant = await createAssistant({
+    name: 'sqlrooms-ai',
+    modelProvider: 'openai',
```
ilyabo: The model provider could also be configurable

examples/sqlrooms-ai/src/store.ts
```diff
@@ -120,6 +121,14 @@ export const {projectStore, useProjectStore} = createProjectStore<
             structuredOutputs: true,
           });
         },
+        getApiKey: () => {
```
ilyabo: Looks like we don't need `createModel` anymore. So we should redesign this interface. I think we need to pass: 

1. A list of models we allow users to choose from which can be from multiple providers
2. API keys for all providers, maybe like this: `{ openai: 'ek....', anthropic: '....', }`

ilyabo (commented):
thanks Xun, great to see this coming together!
I left a few comments

ilyabo:
I noticed something: I tried to ask a few questions, at some point it did run a few queries but didn't show any result. Why could that be?

image

```
{
  "id": "iz5ao0ow23y44ttoe7345ybf",
  "prompt": "make the timestamps and coordinates more human readable ",
  "toolResults": [],
  "toolCalls": [
    {
      "type": "tool-call",
      "toolCallId": "call_M6sVTqsfxWnDbZQlRODAjMlH",
      "toolName": "query",
      "args": {
        "type": "query",
        "sqlQuery": "SELECT DateTime, Latitude, Longitude, Magnitude FROM earthquakes ORDER BY Magnitude DESC LIMIT 5;",
        "reasoning": "To convert the Unix timestamps to a human-readable format, I will use the appropriate SQL function to format the DateTime."
      }
    },
    {
      "type": "tool-call",
      "toolCallId": "call_lK6oJpgLVbGsB0fBBQ5vjRnG",
      "toolName": "query",
      "args": {
        "type": "query",
        "sqlQuery": "SELECT TO_TIMESTAMP(DateTime / 1000) AS HumanReadableDateTime, Latitude, Longitude, Magnitude FROM earthquakes ORDER BY Magnitude DESC LIMIT 5;",
        "reasoning": "To convert the Unix timestamps to a human-readable format, I will use the TO_TIMESTAMP function."
      }
    },
    {
      "type": "tool-call",
      "toolCallId": "call_C4CM0C1BX8g18G5qVipvmaVt",
      "toolName": "query",
      "args": {
        "type": "query",
        "sqlQuery": "SELECT DateTime, Latitude, Longitude, Magnitude FROM earthquakes ORDER BY Magnitude DESC LIMIT 5;",
        "reasoning": "To get the original data for comparison."
      }
    }
  ],
  "analysis": "",
  "isCompleted": true
}
```

ibgreen:
> I noticed something: I tried to ask a few questions, at some point it did run a few queries but didn't show any result. Why could that be?

Yes this is big impact area of improvement. Better error handling, explaining when things failed, how they failed, simply concluding that the failed. Anything is better than silent failures.

lixun910:
Wow, Thanks for your quick response! 😁 Great comments! Yes, I am still working on this draft. There is something I need to change in openassistant/core package as well. Will address the comments and keep you updated.

lixun910:
@ilyabo 

sqlroom-2

lixun910:
Screenshot 2025-02-21 at 10 54 38 PM
Screenshot 2025-02-21 at 10 54 51 PM

ilyabo:
I  ❤️ the ability to view the "Query result", wanted to do the same thing. 
Maybe it should be opening up in a modal, like the one when clicking on a table in the sidebar. For example, @ibgreen is actually using the AI chat in a sidebar panel, so there's little horizontal space.

image

examples/sqlrooms-ai/src/store.ts
lixun910: Yes, we can definitely do that 👍

lixun910:
@ilyabo query table modal is used. Please help to re-review :) 

I will create another branch to add more model providers

packages/ai/src/QueryResult.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * A component that displays a paginated data table using Apache Arrow data.
+ *
+ * @component
+ * @param {Object} props - The component props
+ * @param {ArrowTable} props.arrowTable - The Apache Arrow table containing the data to display
+ * @returns {JSX.Element} A paginated data table component
+ */
+function QueryResultTable({arrowTable}: {arrowTable: ArrowTable}) {
```
ilyabo: I think we shouldn't reimplement `QueryDataTable` and `DataTableModal`, could we try to reuse them? If any changes are necessary, let's make the changes. They currently don't accept arrowTable, but I think we can just pass the query, this way we don't need to keep all the past query results.

lixun910: Yes, I agree. I am not confident to modify the DataTableModal (to break any other parts :). But I can create another PR to get this done after getting more familiar with the code.

ilyabo:
@lixun910 you should rename the PR to `feat: switch to openassistant/core for AI module` for the CI to pass. It uses [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) which are then grouped and included in the changelogs.

PR #14, opened 2025-02-24
[Draft] Add OpenAssistant/Kepler.gl AI plugin

lixun910:
ToDo:
- [x] create map from table
- [ ] create map from query result
- [ ] migrate function tools from kepler.gl ai assistant

Note:
- [ ] Merge after https://github.com/sqlrooms/sqlrooms/pull/11

sqlroom-kepler


ilyabo:
It's mind-blowing!  Only I would really want to avoid `@sqlrooms/ai` depending on `kepler.gl`. Could this be an optional tool we provide when configuring/creating the `AiSlice`?

ilyabo:
and/or we could consider adding a new module, e.g. `@sqlrooms/ai-kepler` to place the code for these tools

PR #15, opened 2025-02-24
feat: Introducing cosmos slice

ilyabo:


.github/workflows/pr-checks.yml
```diff
@@ -45,34 +45,37 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '22'
+          node-version: 22
 
       - name: Setup pnpm
-        uses: pnpm/action-setup@v3
+        uses: pnpm/action-setup@v2
```
lixun910: why downgrade to v2? it seems v4 is latest?

examples/cosmos/src/components/mammals-graph/MammalsGraph.tsx
```diff
@@ -53,12 +61,12 @@ export const MammalsGraph: FC = () => {
...
       linkIndexes={graphData.linkIndexes}
       linkColors={graphData.linkColors}
-      getPointTooltip={(index) => String(graphData.nodes[index])}
+      renderPointTooltip={(index) => String(graphData.nodes[index])}
     >
       <CosmosGraphControls>
         <DownloadButton nodes={graphData.nodes} />
       </CosmosGraphControls>
-      <CosmosSimulationControls />
+      <CosmosSimulationControls className="absolute top-1 right-1" />
```
lixun910, line 69: Nit: Maybe allow additional className for controls?

lixun910 (approved):

.github/workflows/pr-checks.yml
ilyabo: There was some weird issue, apparently with the pnpm cache, I was trying everything to fix, still not sure what actually worked.

PR #25, opened 2025-03-02
feat: Support for Ai Sessions

ilyabo:
- refactor: Project config moved to top level of ProjectStore (to simplify persistence)
- feat: Ai example uses persistence
- feat: Support for Ai Sessions
- feat: UI improvements for Ai
- feat: better markdown rendering

image

Issue #20, opened 2025-02-26
[Enh] add model provider selector in AI example


lixun910:
Add more model provider for AI example

- add model provider selector
- update model selector
- add base url input for (Ollama)

lixun910:
<img width="554" alt="Image" src="https://github.com/user-attachments/assets/8f0484a5-4c14-4c2f-81ef-b6aea1e6829f" />

PR #27, opened 2025-03-05
feat: support persistent ai messages

lixun910:
**NOTE** this PR is for https://github.com/sqlrooms/sqlrooms/pull/25 (branch: project-config-moved-to-top-level)

- support persistent messages(analysisResults)
  - load persistent messages
  - allow delete analysisResult (click 'trash' icon when mouse hover the conversation)
- extend `tool` vercel AI SDK to support:  `execute`,`context` and `component`
  - `execute` needs to return
    - llmResult: the result send back to LLM (no raw data)
    - additionalData: the data will be used by `component` and next `tool`
  - `context`
    - provide e.g. runtime or async data for `execute`
    - `execute` can access `context` via `options.context` 
  - `component`
    - use `additionalData` to render a React component for this `tool`
- sanitize the sql returned by LLM
- ~~support streaming tool calls~~ (support multiple tools and text streaming, so streaming tool calls not making sense anymore)
- support more models https://github.com/sqlrooms/sqlrooms/issues/20
  - OpenAI / Anthropic / Google /Deepseek
  - token saved with provider in localstorage
  - ~~ollama needs input box for base url~~ (in another pr)
- simplify streaming callback function (~~onStepFinish()~~, no longer needed)
- simplify persistent data structure (see below)

The basic data structure of the AI package is:

```ts
    ai: {
      sessions: [
        {
          id: defaultSessionId,
          name: 'Default Session',
          modelProvider: 'openai',
          model: 'gpt-4o-mini',
          analysisResults: [],
          createdAt: new Date(),
        },
      ],
      currentSessionId: defaultSessionId,
    },
```

Each session has a list of `analysisResults` which is an array of `AnalysisResult`.Each `AnalysisResult` contains the following properties:

- `id`: The unique identifier for the analysis result
- `prompt`: The user prompt that was used to generate the analysis result
- `streamMessage`: The stream message from the LLM
- `errorMessage`: The error message from the LLM
- `isCompleted`: Whether the analysis result has been completed

For each user prompt, the LLM will run multiple tools (e.g. `query`, `chart`) and return the result as the `streamMessage`


packages/ai/src/AnalysisResultsContainer.tsx
```diff
@@ -8,16 +8,43 @@ import {
...
+  const [isLoading, setIsLoading] = React.useState(true);
+
+  // TODO: need to wait the datasets to be loaded before rendering the persisted analysis results
+  React.useEffect(() => {
+    const timer = setTimeout(() => {
+      setIsLoading(false);
+    }, 2000);
+
+    return () => clearTimeout(timer);
+  }, []);
```
lixun910: @ilyabo Is there anyway to know the datasets have already been loaded? I want to render the AnalysisResults after datasets been loaded into duckdb. Thanks!

ilyabo: currently you can do this to check whether the data has been loaded (on initialization):
```
const isDataAvailable = useProjectStore(state => state.project.isDataAvailable)
```
or for a particular table:
```
  const isTableReady = useProjectStore((state) =>
    Boolean(state.project.tables.find((t) => t.tableName === 'mammals')),
  );
  ```

PR #32, opened 2025-03-30
feat: Kepler integration

ilyabo:


packages/kepler/src/components/KeplerMapContainer.tsx
```diff
@@ -37,12 +39,19 @@ export const KeplerMapContainer: FC<{
   const dispatchAction = useStoreWithKepler(
     (state) => state.kepler.dispatchAction,
   );
-  const keplerState = useStoreWithKepler((state) => state.kepler.map);
+  const forwardToDispatch = useMemo(
+    () => forwardTo(mapId, dispatchAction),
```
ilyabo: @heshan0131 I think we should encapsulate the forwarding within the `dispatchAction` which we expose to clients so that they don't have to think about it. Maybe we can just add `mapId` as  a parameter to  [this function](https://github.com/sqlrooms/sqlrooms/pull/32/commits/9c4fae1ba900c651b604b77976537e3d495e4db8#diff-b84ff1dd70ac6f5bdb9bbf1cdd59adcb695611ff6b21ff9aef1b40ae75e90021R114). We could also consider automatically adding `mapId: currentMapId` to each action when it is dispatched.

PR #46, opened 2025-04-22
feat: support dropdown actions in table card

heshan0131:
- Allow passing menurender to `TableCard`
Screenshot 2025-04-22 at 12 12 34 PM

### DIFF
--- a/packages/project-builder/src/data-sources/TableCard.tsx
+++ b/packages/project-builder/src/data-sources/TableCard.tsx
@@ -5,7 +5,6 @@ import {
   cn,
   DropdownMenu,
   DropdownMenuContent,
-  DropdownMenuItem,
   DropdownMenuTrigger,
   Tooltip,
   TooltipContent,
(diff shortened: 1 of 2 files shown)


ilyabo:
Shan, following our Slack exchange, I thought about this a bit more.  I think we could do either 

Option 1: "prop-based slots" 

```jsx
<TableCard 
  header={<CustomHeader />}
  footer={<CustomFooter />}
  menuRenderer={(table) => (
  <> {/* allows for items customization depending on the table /*}
    <DropdownMenuItem onClick={() => doSomething(table)}>{`Add table ${table.name}`}</DropdownMenuItem>
    <DropdownMenuSeparator />
    <DropdownMenuItem onClick={() => remove(table)}>Remove</DropdownMenuItem>
  </>
)} />
```

or Option 2: "compound components"
```jsx
<TableCard>
  <TableCard.Header>
    <CustomHeader />
  </TableCard.Header>
  <TableCard.Menu>
    <TableMenu />
  </TableCard.Menu>
  <TableCard.Content>
    <Content />
  </TableCard.Content>
  <TableCard.Footer>
    <Footer />
  </TableCard.Footer>
</TableCard>
```

Option 2 is most flexible but might be an overkill for simple cases like this which don't require **_layout_** customization or the ability to reuse the parts independently. I think Option 2 is very appropriate for #45, but for the table card Option 1 is good enough and is much less verbose, so I would go with Option 1 here.

ilyabo (approved):

PR #53, opened 2025-05-14
[Bug] Fix dispatch middleware

heshan0131:
- Fix middleware dispatch not clearning Task

### DIFF
--- a/packages/kepler/src/KeplerSlice.ts
+++ b/packages/kepler/src/KeplerSlice.ts
@@ -6,43 +6,40 @@ import {
   requestMapStyles,
   wrapTo,
 } from '@kepler.gl/actions';
-import {
-  ALL_FIELD_TYPES,
-  VectorTileDatasetMetadata
-} from '@kepler.gl/constants';
+import {ALL_FIELD_TYPES, VectorTileDatasetMetadata} from '@kepler.gl/constants';
 import {
   constructST_asWKBQuery,
   getDuckDBColumnTypes,
   getDuckDBColumnTypesMap,
   getGeometryColumns,
   restoreGeoarrowMetadata,
-  setGeoArrowWKBExtension
+  setGeoArrowWKBExtension,
 } from '@kepler.gl/duckdb';
-import { arrowSchemaToFields } from '@kepler.gl/processors';
-import {
-  keplerGlReducer,
-  KeplerGlState,
-  MapStyle
-} from '@kepler.gl/reducers';
+import {arrowSchemaToFields} from '@kepler.gl/processors';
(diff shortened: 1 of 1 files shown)


packages/kepler/src/KeplerSlice.ts
```diff
@@ -207,74 +198,105 @@ export function createKeplerSlice<
...
+            set((state: KeplerSliceState<PC>) => ({
+              ...state,
+              kepler: {
+                ...state.kepler,
+                map: keplerReducer(state.kepler.map, action),
+              },
+            }));
+
+            // Call onAction if it's defined
+            get().kepler.onAction?.(mapId, action);
```
heshan0131: @ilyabo this will no longer work. Which store is calling kepler.onAction? and what's this for?

ilyabo: It's not used at the moment, but I though might be good for clients to have a possibility to listen and react to Kepler actions. I added it after you and I had discussed a potential fix for the loading issues.

Why won't it work anymore?

packages/kepler/src/KeplerSlice.ts
```diff
@@ -207,74 +198,105 @@ export function createKeplerSlice<
...
+            middleware(middlewareAPI),
+          );
+          wrapDispatch = compose<Dispatch>(...chain)(storeDispatch);
           set({
             kepler: {
               ...get().kepler,
               map: keplerInitialState,
-              dispatchAction: dispatchWithMiddleware,
+              dispatchAction: (mapId, action) =>
+                wrapDispatch(wrapTo(mapId, action)),
```
ilyabo: Nice, so wrapTo adds `mapId` to each action's payload, right?

ilyabo (approved):

packages/kepler/src/KeplerSlice.ts
ilyabo: so it would be called if a client adds an `onAction` implementation to `.kepler` when setting up the store

packages/kepler/src/KeplerSlice.ts
heshan0131: Yes, so only need to pass `action` to `wrapDispatch`, because the middleware doesn't take `(mapId, action)` as input. We need to pass a regular `dispatch` to middlewares for chaining

packages/kepler/src/KeplerSlice.ts
heshan0131: because `sotreDispatch` no longer takes `mapId` as arguments. Would need to get `mapId` from `action.payload.meta._id_`

PR #62, opened 2025-05-20
fix: monaco editor force rerender when update theme

heshan0131:


packages/monaco-editor/src/components/MonacoEditor.tsx
```diff
@@ -205,6 +206,9 @@ export const MonacoEditor: React.FC<MonacoEditorProps> = ({
           isDark ? 'sqlrooms-dark' : 'sqlrooms-light',
         );
       }
+
+      // Force re-render to apply the new theme
+      setRenderKey((key) => key + 1);
```
ilyabo, line 211: do we really need the counter here? I think it should be enough to just use smth like 
```
key={isDark?'dark':'light'}
```
for simplicity

ilyabo (approved):

packages/monaco-editor/src/components/MonacoEditor.tsx
heshan0131: What to use as key doesn't matter, it can be 'dark' or 'light' or key + 1, but we still need to call `seRenderKey` here to trigger a state change, which will force rerender

PR #63, opened 2025-05-23
feat: create kepler add data modal, s3 file browser, s3 credential form

heshan0131:
-  in kepler module add KeplerAddDataDialog with 3 methods: upload | tileset | s3
- in kepler module add S3Browser 
- in sqlrooms/s3-browser module add S3CredentialForm
- move s3 utils out of sqlrooms/s3-browser into a seperate sqlrooms/s3 module to be load on the server
- remove react 19 requirements 

Screenshot 2025-05-30 at 12 33 10 PM
Screenshot 2025-05-30 at 12 33 14 PM
Screenshot 2025-05-30 at 12 34 07 PM
Screenshot 2025-06-02 at 9 25 22 PM


packages/duckdb/src/DuckDbSlice.ts
```diff
@@ -320,7 +320,10 @@ export function createDuckDbSlice({
           return getColValAsNumber(res) > 0;
         },
 
-        async dropTable(tableName, options): Promise<void> {
+        async dropTable(
+          tableName: string,
+          options?: {schema?: string; database?: string},
```
ilyabo, line 325: nit: I think this shouldn't be necessary if we have the types in the interface

packages/s3/README.md
```diff
@@ -0,0 +1,73 @@
+This package is part of the SQLRooms framework.
```
ilyabo, line 1: Shan, sorry, I think you might've told me, but I forgot. Is there a reason for having a separate package for the utils if they are intended for use with `s3-browser` anyways? Otherwise, I think, it would be less confusing if we could keep it in one package.

packages/kepler/src/components/KeplerAddDataDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+  TabsTrigger,
+} from '@sqlrooms/ui';
+import {LoadTileSetFactory, Icons} from '@kepler.gl/components';
+import {FileDropInput} from './FileDropInput';
+import {KeplerInjector} from './KeplerInjector';
+import {KeplerProvider} from './KeplerProvider';
+import {S3Browser, S3BrowserProps} from './S3Browser';
+import {useIntl} from 'react-intl';
+
+const ACCEPTED_FORMATS = [
```
ilyabo: Maybe this should be configurable via `createKeplerSlice` (with a default value)? In that case we could pass the same list as the one we show in the sidebar

packages/kepler/src/components/KeplerAddDataDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+  );
+  const onLoadS3Files = useCallback(
+    async (...args: Parameters<S3BrowserProps['loadS3Files']>) => {
+      await loadS3Files(...args);
+      addDataModal.onClose();
+    },
+    [loadS3Files, addDataModal],
+  );
+  return (
+    <KeplerProvider mapId={''}>
```
ilyabo, line 117: Maybe add a comment why `mapId` isn't necessary here?

packages/kepler/src/components/S3Browser.tsx
```diff
@@ -0,0 +1,181 @@
...
+    s3Config: S3Config;
+    prefix: string;
+    files: string[];
+  }) => Promise<void>;
+  s3: S3State['s3'];
+  saveS3Connection: (s3Config: S3Config) => Promise<void>;
+  loadS3Connections: () => Promise<S3Connection[]>;
+  deleteS3Connection: (id: string) => Promise<void>;
+};
+export const S3Browser = ({
```
ilyabo: Maybe name it `KeplerS3Browser` if it's specific to kepler?

packages/s3-browser/src/S3Slice.ts
```diff
@@ -0,0 +1,43 @@
...
+    // current in credential form
+    currentS3Config: S3Config | null;
+    // Credential management
+    setCurrentS3Config: (s3Config: S3Config) => void;
+    clearCurrentS3Config: () => void;
+  };
+};
+
+// Create the store
+export function createS3Slice<
```
ilyabo: I think we should probably use `S3BrowserState`, `S3BrowserSlice` , `S3BrowserConfig`, `createS3BrowserSlice` to align with the name of the package and other slices, especially, if we need to have `@sqlrooms/s3` as a separate package

packages/s3-browser/src/S3FileBrowser.tsx
```diff
@@ -14,11 +15,37 @@ import {
...
 } from '@sqlrooms/ui';
-import {Undo2Icon, FolderIcon} from 'lucide-react';
+import {S3FileOrDirectory} from '@sqlrooms/s3';
+import {Undo2Icon, FolderIcon, Search} from 'lucide-react';
 import {formatBytes, formatTimeRelative} from '@sqlrooms/utils';
-import {FC, useCallback, useEffect, useMemo} from 'react';
-import {S3FileOrDirectory} from './S3FileOrDirectory';
+import {FC, useCallback, useEffect, useMemo, useState} from 'react';
 
+function renderFileName(
```
ilyabo, line 23: Maybe consider using `SearchTextHighlight`  from the [previous Condor incarnation](https://github.com/foursquare/sql-workbooks/blob/main/apps/condor/src/components/explorer/nodes/search-text-highlight.tsx#L16)?

packages/s3-browser/src/S3FileBrowser.tsx
```diff
@@ -79,11 +107,23 @@ const S3FileBrowser: FC<{
     onCanConfirmChange,
     onChangeSelectedFiles,
     onChangeSelectedDirectory,
+    renderFileActions,
   } = props;
 
   useEffect(() => {
     onCanConfirmChange(Boolean(selectedFiles?.length));
   }, [selectedFiles, onCanConfirmChange]);
+  const [searchQuery, setSearchQuery] = useState('');
```
ilyabo, line 116: Would it make sense to move this state and logic to `S3BrowserSlice`?

packages/s3-browser/README.md
```diff
@@ -24,6 +24,117 @@ yarn add @sqlrooms/s3-browser
...
+
+export const S3Browser = ({
+  listS3Files,
+  s3,
+  loadS3Files,
+  saveS3Connection,
+  loadS3Connections,
+  deleteS3Connection,
+}: S3BrowserProps) => {
+  const [isConnecting, setIsConnecting] = useState(false);
```
ilyabo, line 58: I think it would make sense to keep all this state and logic in `S3BrowserSlice`, then the usage would be more straightforward and we wouldn't need to pass so many props to the components.

packages/s3/src/S3Connection.ts
```diff
@@ -0,0 +1,26 @@
...
+  secretAccessKey: z.string().min(1, 'Secret Access Key is required'),
+  region: z.string().min(1, 'Region is required'),
+  bucket: z.string().min(1, 'Bucket name is required'),
+  name: z.string().optional(), // Optional for saving
+  sessionToken: z.string().optional(), // Optional for temporary credentials
+});
+
+export type S3Config = z.infer<typeof S3ConfigSchema>;
+
+export const S3ConnectionSchema = z.object({
```
ilyabo: I think we shouldn't use the terms "connect" and "connection", since no persistent connection is established in this case. I would just go with "Credential", "S3 Credentials",  "Select a saved credential". [ChatGPT agrees](https://chatgpt.com/share/683ed010-9028-8001-b415-d4882263851a)

ilyabo (approved):
looks great! 👍 
I left a few comments, but I'm fine with addressing them later

packages/s3-browser/README.md
```diff
@@ -60,33 +171,107 @@ function MyS3Browser() {
...
-async function listFiles() {
-  const files = await listFilesAndDirectoriesWithPrefix(
-    s3Client,
-    'my-bucket',
-    'path/to/directory',
+import {S3CredentialForm} from '@sqlrooms/s3-browser';
+import {S3Config, S3Connection} from '@sqlrooms/s3';
+
+function MyS3ConnectionManager() {
+  const handleConnect = async (credentials: S3Config) => {
```
ilyabo, line 181: Same here, I think all this logic could be encapsulated within the slice.  For loading/saving I think it would be better to just add `initialCredentials` and `onSaveCredentials` props to `createS3BrowserSlice`.

packages/s3/README.md
heshan0131: The aws commands should only be called on the server side. And we don't want to bundle all the ui components there

packages/s3/src/S3Connection.ts
heshan0131: Sure, but I have to say this prompt is very leading, by mentioning credentials and suggesting 'connection isn't the right name right?'

packages/s3-browser/README.md
heshan0131: I think it's fine either way. right now by embeding the logic inside the component. it can be used without zustand or redux. User can choose their own async mgmt library. ie. I can use thisin kepler.gl. But if moved the async logic to a store, it can only be used with zustand

PR #82, opened 2025-06-24
feat: Upgrade open assistant

ilyabo:


examples/ai/package.json
```diff
@@ -28,6 +28,7 @@
     "ai": "^4.3.16",
     "immer": "^10.1.1",
     "lucide-react": "^0.474.0",
+    "ollama-ai-provider-v2": "^0.0.5",
```
lixun910, line 31: Ollama-ai-provider is not actively maintained, switch to ollama-ai-provider-v2	https://github.com/vercel/ai/issues/4700

examples/ai/src/example-sessions.json
```diff
@@ -10,636 +10,324 @@
           "id": "hq9x8r95oennvhst8o6ollpm",
           "prompt": "draw time series",
           "streamMessage": {
-            "toolCallMessages": [
+            "parts": [
```
ilyabo: @lixun910 would it be difficult to provide [`z.preprocess`](https://v3.zod.dev/?id=preprocess) to all the changed schemas so that the existing saved configs would be automatically migrated?

ilyabo: it could looks smth like this, only the changes from `StreamMessageSchema` to `StreamMessagePartSchema` seem to be quite substantial:

```tsx
export const AnalysisResultSchema = z.object({
  id: z.string().cuid2(),
  prompt: z.string(),
  streamMessage: z.preprocess(
    (data) => {
      if (typeof data !== 'object' || data === null || !('parts' in data)) {
        const oldMessage = data as any;
        return {parts: oldMessage.toolCallMessages.map(msg => ({
          ...
        }))};
      }
      return data;
    },
    z.object({
      parts: z.array(StreamMessagePartSchema).optional(),
    }),
  ),
  errorMessage: ErrorMessageSchema.optional(),
  isCompleted: z.boolean(),
});
```

lixun910: Oh, looks like the z.preprocess() is a very useful function for migration. Will try it. Thanks!

lixun910: Good to learn this function! Works perfectly for migration 👍

PR #116, opened 2025-08-19
feat: support connect to proxy llm server

lixun910:
- Allow specify LLM Proxy server
  - If proxy server is provided, no need to provide apiToken

e.g. connecting to a locally run liteLLM proxy server at port 4000
Screenshot 2025-08-19 at 4 46 00 PM

### DIFF
--- a/examples/ai/src/models.ts
+++ b/examples/ai/src/models.ts
@@ -4,6 +4,15 @@ export const DEFAULT_MODEL = 'gpt-4o-mini';
 export const OLLAMA_DEFAULT_BASE_URL = 'http://localhost:11434/api';
 export const CUSTOM_MODEL_NAME = 'custom';
 
+// Default base URLs for each provider
+export const PROVIDER_DEFAULT_BASE_URLS = {
+  openai: 'https://api.openai.com/v1',
+  anthropic: 'https://api.anthropic.com/v1',
+  google: 'https://generativelanguage.googleapis.com/v1beta',
+  deepseek: 'https://api.deepseek.com/v1',
+  ollama: OLLAMA_DEFAULT_BASE_URL,
+} as const;
+
 export const LLM_MODELS = [
   {
     name: 'openai',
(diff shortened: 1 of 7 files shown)


ilyabo (approved):

packages/ai/src/AiSlice.ts
```diff
@@ -83,7 +83,7 @@ export function createAiSlice<PC extends BaseRoomConfig & AiSliceConfig>({
   customTools = {},
   getInstructions,
 }: {
-  getApiKey: (modelProvider: string) => string;
+  getApiKey?: (modelProvider: string) => string;
```
ilyabo: I think it'd be good to have baseUrl as one of the props. Could we add a union type here? Something like:

```ts
 {... /* rest of the props */ } &
 ({baseUrl: string}  |
  {getApiKey: (modelProvider: string) => string;})
```

examples/ai/src/models.ts
```diff
@@ -4,6 +4,15 @@ export const DEFAULT_MODEL = 'gpt-4o-mini';
 export const OLLAMA_DEFAULT_BASE_URL = 'http://localhost:11434/api';
 export const CUSTOM_MODEL_NAME = 'custom';
 
+// Default base URLs for each provider
+export const PROVIDER_DEFAULT_BASE_URLS = {
```
ilyabo, line 8: Do we really need these? Aren't these URLs bundled into the providers libraries' already, so that if we don't pass custom ones, the default ones will be used?

ilyabo: In particular, for the example app, maybe we should hide this input field. It's rather for advanced use.

lixun910: Yes, we need these for the feature that if user clears the baseUrl, the original baseUrl will be shown (so users don't need to remember them :).
Since the provider packages are dynamic loaded, we have to manually code here.

PR #132, opened 2025-09-06
feat: add ai-chatui package

lixun910:


examples/ai/src/components/MainView.tsx
```diff
@@ -1,184 +1,86 @@
-import {Input, SkeletonPane} from '@sqlrooms/ui';
-import {KeyIcon, ServerIcon, CpuIcon} from 'lucide-react';
+import {useState} from 'react';
+import {Button, SkeletonPane} from '@sqlrooms/ui';
+import {Settings} from 'lucide-react';
 import {
+  AiConfigPanel,
```
ilyabo: Let's call it `AiSettingsPanel` to differentiate from the `config` we store as part of the app state.

ilyabo (commented):
I think we should use the term "settings" consistently for all the names related to the model/provider settings which to avoid confusion with the config saved as part of the state, including:

- `AiConfigSlice` -> `AiSettingsSlice`
- `config` directory -> `settings`
- `AiModelsConfig` -> `AiModelsSettings` 
- `AiProvidersConfig` -> `AiProvidersSettings`
...

packages/ai/src/AiSlice.ts
```diff
@@ -339,16 +312,18 @@ export function createAiSlice<PC extends BaseRoomConfig & AiSliceConfig>(
...
             await runAnalysis({
               tableSchemas: get().db.tables,
-              modelProvider: currentSession.modelProvider || 'openai',
+              modelProvider: currentSession.modelProvider || defaultProvider,
               model: currentSession.model || defaultModel,
-              customModelName: currentSession.customModelName,
               apiKey:
-                getApiKey?.(currentSession.modelProvider || 'openai') || '',
-              baseUrl: currentSession.baseUrl || baseUrl,
+                getApiKey?.(currentSession.modelProvider || defaultProvider) ||
```
ilyabo: Shouldn't this line go into to `getApiKeyFromSettings()` so that we can call `getApiKeyFromSettings` from some other place and have the same behavior?

PR #137, opened 2025-09-10
feat: Ai assistant read only mode

ilyabo:


lixun910 (approved):
LGTM 👍 

One thing is that SUMMARIZE lastStatement will need to do an extra query of lastStatement. When the sql of lastStatement is complex and slow, it might impact the performance. This also happens when checking the query table returned by the tool, and the same query will be executed again.

So, probably, what we can do is caching the query results (with size of the first page of query table), we can then run SUMMARIZE using this cached result, and boost the performance of initial rendering when user click the table icon to check the query results from the tool.

PR #153, opened 2025-10-01
refactor: Separate aiSettings config

ilyabo:
- AiSettings config config separated from RoomConfig to make it easier to persist separately and to simplify typing (`state.aiSettings.config` instead of `state.config.aiSettings`)
- Added missing namespacing for aiSettings slice


lixun910:
one quick question: should we do the same for state.config.ai.sessions?

PR #165, opened 2025-10-14
fix: missing dependencies and tailwind config errors

lixun910:
- add some missing dependencies from some packages 
- fix tailwind config (probably misconfigured by llm)

It is also interesting that github actions passed build check. I think maybe caused by the cache?
```yml
      - name: Setup Node
        uses: actions/setup-node@v4
        with:
          node-version: 22
          cache: 'pnpm'
```

### DIFF
--- a/examples/ai-core/tailwind.config.ts
+++ b/examples/ai-core/tailwind.config.ts
@@ -7,7 +7,8 @@ const config = {
     'src/**/*.{ts,tsx}',
     // If you make a precise list of packages used, instead of @sqlrooms/*,
     // it would help Vite start faster in dev mode
-    '**/node_modules/@sqlrooms/*/dist/**/*.js',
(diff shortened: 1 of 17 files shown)


examples/ai-core/tailwind.config.ts
```diff
@@ -8,6 +8,7 @@ const config = {
     // If you make a precise list of packages used, instead of @sqlrooms/*,
     // it would help Vite start faster in dev mode
     '{./,../../}node_modules/@sqlrooms/*/dist/**/*.js',
```
ilyabo, line 10: maybe we should just try to change it to everywhere?
```
'**/node_modules/@sqlrooms/*/dist/**/*.js'
```

ilyabo:
closing, as I think this isn't relevant anymore

PR #167, opened 2025-10-23
feat: RAG Python embedding preparation and querying modules

ilyabo:
image


stprnvsh:
I tried out a version where we can take the schema from the wasm-duckdb, create embeddings directly using an api and save the embeddings as a separate database/knowledge base that the llm can use for getting the necessary schema information by using a hybrid search before curating the sql query 

is there any reason for choosing a python server and not doing it all on the browser side?

ilyabo:
@stprnvsh I was able to take it further, and updated the [Python module](https://pypi.org/project/sqlrooms-rag/). There's now an [ai-rag example](https://github.com/sqlrooms/sqlrooms/tree/main/examples/ai-rag) which can work with a prepared database. What's still missing, is adding documents on-the-fly (I think that might the use-case you were describing, right?). If you could extend the [RagSlice](https://github.com/sqlrooms/sqlrooms/blob/main/packages/ai-rag/src/RagSlice.ts) so that we can add documents at runtime, at least embeddings, not necessarily for the hybrid search, that'd be really cool, I think.

PR #178, opened 2025-11-02
feat: fix height for prompt textarea box

lixun910:
The height of prompt textarea will grow with the input text. This PR tries to make the prompt textarea with fix height. Not sure if "Get max-height constraint from CSS" is the right way to do it. @albatross97 can you help to take a look? Thanks!


ilyabo:
I think it should be enough to just set `max-height`, it should override height, then we don't need all this complex JS logic. I would set it to smth like `max-height: min(300px,40vh)`, that should be reasonable

PR #205, opened 2025-11-25
fix: Layer panel title width

albatross97:
Auto-expand Layer panel title when there are enough horizontal space

Screenshot 2025-11-25 at 11 18 47 AM


ilyabo:
@albatross97 Note, that there's no more `kepler-integration` branch, so I recreated the PR https://github.com/sqlrooms/sqlrooms/pull/209

PR #219, opened 2025-12-02
feat: Multi-statement create table from query

ilyabo:
- Multi-statement create table from query
- Added new duckdb-node module
- Added new duckdb-core module
- Added tests for `DuckDBSlice`

- Improved `CreateTableModal`:

image

### DIFF
--- a/examples/kepler/src/main.tsx
+++ b/examples/kepler/src/main.tsx
@@ -4,6 +4,11 @@ import {createRoot} from 'react-dom/client';
 import {Room} from './room';
 import './index.css';
 
+import {configureMonacoLoader} from '@sqlrooms/monaco-editor';
+import * as monaco from 'monaco-editor';
+import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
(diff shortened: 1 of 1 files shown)


packages/duckdb/src/DuckDbSlice.ts
```diff
@@ -278,33 +294,87 @@ export function createDuckDbSlice({
...
+                })
+              : baseQualifiedName;
+
             const connector = await get().db.getConnector();
 
-            const statements = splitSqlStatements(query);
-            if (statements.length !== 1) {
-              throw new Error('Query must contain exactly one statement');
+            const {precedingStatements, lastStatement} =
+              separateLastStatement(query);
```
lixun910, line 341: Great! Thanks, @ilyabo! 

qq: are we assuming last statement is SELECT statement? is it necessary to find SELECT statement from query? Thanks!

ilyabo: Yes, I think we can assume it should be the last one.
image

PR #220, opened 2025-12-02
feat: Extract scrollable tabs into a component

albatross97:


ilyabo:
Thanks @albatross97! Looking good!
I'll try a few refactorings to make it more composable.
Also, I think it's cleaner instead of `openedTabs` and  `closedTabs` to have `tabs` (all), and `openedTabIds`.

ilyabo (approved):

PR #262, opened 2025-12-11
fix: Prevent infinite rerender in AI AnalysisResultsContainer

ilyabo:
Prevents infinite render update loop in AI AnalysisResultsContainer when there are no sessions and messages due to returning new empty array from state selectors


lixun910 (approved):
Thank you @ilyabo !

PR #267, opened 2025-12-16
fix: agent rendering and add abortSignal to query tool

lixun910:
- fix rendering tool execution in agent
- add abortSignal to query tool and createTableFromQuery

### DIFF
--- a/packages/layout/src/LayoutSlice.ts
+++ b/packages/layout/src/LayoutSlice.ts
@@ -15,7 +15,7 @@ export type RoomPanelInfo = {
   title?: string;
   icon?: React.ComponentType<{className?: string}>;
   component: React.ComponentType;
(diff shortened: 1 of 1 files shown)


packages/layout/src/LayoutSlice.ts
```diff
@@ -15,7 +15,7 @@ export type RoomPanelInfo = {
   title?: string;
   icon?: React.ComponentType<{className?: string}>;
   component: React.ComponentType;
-  placement: 'sidebar' | 'sidebar-bottom' | 'main' | 'top-bar';
+  placement: 'sidebar' | 'sidebar-bottom' | 'main' | 'main-bottom';
```
ilyabo: was this change intentional?

lixun910: Ah, I will revert it. Thanks!

PR #276, opened 2025-12-24
fix(kepler): timeline slider not showing for trip layer

igorDykhta:
- Fix timeline slider not showing for the trip layer in the kepler.gl module


lixun910 (approved):
Thank you! @igorDykhta

ilyabo (approved):

Issue #278, opened 2026-01-06
Switch sessions while chat in streaming


lixun910:
Issue: If you start a chat in Session A, then switch to Session B while it’s streaming, the conversation starts to be appended to Session B

PR #279, opened 2026-01-06
fix: switch sessions while chat in streaming

lixun910:
Issue: If you start a chat in Session A, then switch to Session B while it’s streaming, the conversation starts to be appended to Session B https://github.com/sqlrooms/sqlrooms/issues/278

Fix:

- the conversation continues to be appended to Session A only.
- Session B stays untouched (ThinkingDots should not be rendered in Session B)
- When you return to Session A, you’ll see the completed run there.

### DIFF
--- a/packages/ai-core/src/components/AnalysisResultsContainer.tsx
+++ b/packages/ai-core/src/components/AnalysisResultsContainer.tsx
@@ -1,4 +1,4 @@
-import {Button, cn, ScrollArea, ScrollBar} from '@sqlrooms/ui';
+import {cn, ScrollArea, ScrollBar} from '@sqlrooms/ui';
 import {ChevronDown} from 'lucide-react';
 import React, {useEffect, useRef} from 'react';
 import {Components} from 'react-markdown';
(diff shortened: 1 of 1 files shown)


packages/ai-core/src/components/AnalysisResultsContainer.tsx
```diff
@@ -22,13 +22,19 @@ export const AnalysisResultsContainer: React.FC<{
   ErrorMessageComponent,
 }) => {
   const isRunningAnalysis = useStoreWithAi((s) => s.ai.isRunningAnalysis);
+  const activeChatSessionId = useStoreWithAi((s) => s.ai.activeChatSessionId);
```
ilyabo: I find it confusing to have both "current" and "active". I think it'd be better if we didn't have to introduce `active` as a new concept.  

Also, what happens if two conversations are running in parallel? I start one chat, then I switch to another one and submit a message, while the first is still thinking?

ilyabo: I think, ideally, messages should somehow keep the ids of the sessions the belong to within them, so that we know where they should be added to.

lixun910: 👍 Agreed. Let me update the fix. 

I don’t think the current design supports parallel conversations or sessions. When switching to another session, the Run button still shows as running, and you have to manually stop it before submitting a new message. We may want to redesign AISlice to properly support parallel conversations.

PR #284, opened 2026-01-08
feat: AI: Support parallel sessions [take2]

lixun910:


REFACTOR_SUMMARY.md
```diff
@@ -0,0 +1,184 @@
...
+**New Methods**:
+- `getSessionAbortController(sessionId)`
+- `setSessionAbortController(sessionId, controller)`
+- `getSessionChatStop(sessionId)`
+- `setSessionChatStop(sessionId, stop)`
+- `getSessionChatSendMessage(sessionId)`
+- `setSessionChatSendMessage(sessionId, sendMessage)`
+- `getSessionAddToolResult(sessionId)`
+- `setSessionAddToolResult(sessionId, addToolResult)`
+- `getSessionAnalysisPrompt(sessionId)`
```
ilyabo: let's get rid of all ..Analysis.. in at least the new functions/props we are adding.
at least `SessionAnalysis` is redundant for sure,  just `Session` should be enough

ilyabo: also the new `isRunningAnalysis` and `analysisPrompt` etc

REFACTOR_SUMMARY.md
```diff
@@ -0,0 +1,188 @@
+# AI-Core Per-Session Chat Refactoring Summary
```
ilyabo: We should probably remove this file, and add a section to `upgrade-guide.md` instead describing the API changes

PR #287, opened 2026-01-12
fix: AI Fix tool error message layout

bdjulbic:
Fix layout in the error message

### DIFF
--- a/packages/ai-core/src/components/tools/ToolErrorMessage.tsx
+++ b/packages/ai-core/src/components/tools/ToolErrorMessage.tsx
@@ -1,4 +1,4 @@
-import { JsonMonacoEditor } from '@sqlrooms/monaco-editor';
+import {JsonMonacoEditor} from '@sqlrooms/monaco-editor';
 import {
   Button,
   Popover,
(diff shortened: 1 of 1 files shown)


packages/ai-core/src/components/tools/ToolErrorMessage.tsx
```diff
@@ -61,33 +61,38 @@ export function ToolErrorMessage(props: ToolErrorMessageProps) {
...
-                  minimap: {enabled: false},
-                  scrollBeyondLastLine: false,
-                  wordWrap: 'on',
-                }}
-              />
-            </div>
-          ) : null}
-        </div>
-      </PopoverContent>
+      {popoverOpen.isOpen ? (
```
bdjulbic, line 64: I moved up this check `{popoverOpen.isOpen ?`
cause during unmounting process, the empty popup component would show up in different part of the screen for a split second

This wasn't noticeable before, until the layout was fixed

packages/ai-core/src/components/tools/ToolErrorMessage.tsx
```diff
@@ -1,12 +1,12 @@
-import {JsonMonacoEditor} from '@sqlrooms/monaco-editor';
+import { JsonMonacoEditor } from '@sqlrooms/monaco-editor';
```
bdjulbic: I ran document reformat, so this change appeared

bdjulbic: This is fixed now

PR #291, opened 2026-01-14
fix: dynamic font class resolver

bdjulbic:
Fix font size customization via property in useArrowDataTable.tsx
And also update DataTablePaginated just so its uniform across sqlrooms

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,8 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml
-# Editor-based HTTP Client requests
-/httpRequests/
-# Datasource local storage ignored files
-/dataSources/
-/dataSources.local.xml


packages/data-table/src/DataTablePaginated.tsx
```diff
@@ -134,7 +135,7 @@ export default function DataTablePaginated<Data extends object>({
               {table.getHeaderGroups().map((headerGroup) => (
                 <TableRow key={headerGroup.id}>
                   <TableHead
-                    className={`bg-background sticky left-0 top-[-1px] z-10 w-auto whitespace-nowrap border-r py-2 text-center`}
+                    className={`bg-background sticky top-[-1px] left-0 z-10 w-auto border-r py-2 text-center whitespace-nowrap`}
```
bdjulbic, line 139: prettifier changes ordering of this classnames,
I assume prettify config got updated since the file was created

ilyabo (approved):
lgtm, thanks!

PR #303, opened 2026-01-19
fix: show chart as inline component in the sidebar

bdjulbic:
Two small UI tweaks:
- Tool error popover: 
don’t show the plain error text when details are present, since the JSON Monaco viewer already displays it.
Screenshot 2026-01-19 at 17 02 58

- AI assistant sidebar: 
render the chart tool result inline in the message flow (excluded via userTools), instead of grouping it inside the collapsible reasoning box.
Screenshot 2026-01-19 at 17 03 07

### DIFF
--- a/packages/ai-core/src/components/AnalysisResult.tsx
+++ b/packages/ai-core/src/components/AnalysisResult.tsx
@@ -22,8 +22,7 @@ type AnalysisResultProps = {
   analysisResult: AnalysisResultSchema;
   enableReasoningBox?: boolean;
   customMarkdownComponents?: Partial<Components>;
-  // TODO: maybe rename to excludeFromGrouping
-  userTools?: string[];
+  excludeFromGrouping?: string[];
   ErrorMessageComponent?: React.ComponentType<ErrorMessageComponentProps>;
 };
 
@@ -41,7 +40,7 @@ export const AnalysisResult: React.FC<AnalysisResultProps> = ({
   analysisResult,
   enableReasoningBox = false,
   customMarkdownComponents,
-  userTools,
+  excludeFromGrouping: userTools,
   ErrorMessageComponent,
 }) => {
   const uiMessages = useStoreWithAi(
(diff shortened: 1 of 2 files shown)


packages/vega/src/VegaChartToolResult.tsx
```diff
@@ -96,7 +96,7 @@ export function VegaChartToolResult({
         onSqlChange={handleSqlChange}
       >
         {/* Chart with actions toolbar */}
-        <div className="relative min-h-[300px]">
```
bdjulbic, line 99: I removed the height constraint 
because its parent has a aspect ratio 16:9 constraint. So it should be fine

Any suggestion where to test this?

packages/ai-core/src/components/AnalysisResult.tsx
```diff
@@ -22,6 +22,7 @@ type AnalysisResultProps = {
   analysisResult: AnalysisResultSchema;
   enableReasoningBox?: boolean;
   customMarkdownComponents?: Partial<Components>;
+  // TODO: maybe rename to excludeFromGrouping
   userTools?: string[];
```
bdjulbic: I would suggest renaming this to more specific name like excludeFromGrouping

The tools listed here aren't grouped later on

ilyabo:
> Tool error popover: don’t show the plain error text when details are present, since the JSON Monaco viewer already displays it.

I'm thinking, maybe we should have two tabs there:  "Error text" showing the error text with `whitespace-prewrap font-mono`, and another one with the "Full tool call", because the error text is difficult to read within JSON

packages/ai-core/src/components/AnalysisResult.tsx
lixun910: Yes, sounds good 👍

lixun910 (approved):
LGTM 👍 

Yes, sounds good changing `userTools` to `excludeFromGrouping`.

PR #308, opened 2026-01-22
feat: improve explain query output in sqleditor

bdjulbic:
Show explain schema as a full height content in the QueryResultsPanel instead as a table row

Screenrecording:
- show regular table dataset, and explain schema
explain

### DIFF
--- a/packages/sql-editor/src/components/QueryResultPanel.tsx
+++ b/packages/sql-editor/src/components/QueryResultPanel.tsx
@@ -94,10 +94,6 @@ export const QueryResultPanel: React.FC<QueryResultPanelProps> = ({
     (s) => s.sqlEditor.queryResultLimitOptions,
   );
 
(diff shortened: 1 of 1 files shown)


packages/sql-editor/src/components/QueryResultPanel.tsx
```diff
@@ -122,14 +158,35 @@ export const QueryResultPanel: React.FC<QueryResultPanelProps> = ({
   }
 
   if (queryResult?.status === 'success') {
-    return (
-      <div
-        className={cn(
-          'relative flex h-full w-full flex-grow flex-col overflow-hidden',
-          className,
```
bdjulbic, line 129: I split the previous long ternary operator result into multiple early returns.

Extracted the container classnames as a const `contentWrapperClassName`  and applied it to the early returns

PR #328, opened 2026-01-27
fix: chart label position and add responsive font size

bdjulbic:
Improve the chart presentation in the ai assistant sidebar,
Label position is improved and also added simple font size mapping based on the container width

chart label font

### DIFF
--- a/packages/vega/src/VegaLiteArrowChart.tsx
+++ b/packages/vega/src/VegaLiteArrowChart.tsx
@@ -127,6 +127,22 @@ const VegaLiteArrowChartBase: React.FC<VegaLiteArrowChartProps> = ({
     options,
   });
 
+  const dimensions = useAspectRatioDimensions({
+    containerRef,
+    width,
+    height,
+    aspectRatio,
+  });
+  const changeDimensions = useCallback(
+    (width: number, height: number) => {
+      embed?.view.width(width).height(height).runAsync();
+    },
+    [embed],
+  );
+  useEffect(() => {
+    changeDimensions(dimensions.width, dimensions.height);
+  }, [changeDimensions, dimensions.width, dimensions.height]);
+
   return (
     <VegaChartContextProvider value={{embed}}>
       <div
@@ -149,12 +165,7 @@ const VegaLiteArrowChartBase: React.FC<VegaLiteArrowChartProps> = ({
               className="overflow-visible"
               asChild
             >
-              <div
-                ref={ref}
-                // Ensure the Vega SVG title isn't clipped when it shifts upward
-                // on small container widths.
-                className="[&_svg]:overflow-visible"
-              />
+              <div ref={ref} className="[&_svg]:overflow-visible" />
             </AspectRatio>
           )
         )}


packages/vega/src/VegaLiteArrowChart.tsx
```diff
@@ -61,6 +61,48 @@ export function makeDefaultVegaLiteOptions(
...
+  let preset: {
+    axisLabel: number;
+    axisTitle: number;
+    legendLabel: number;
+    legendTitle: number;
+    title: number;
+  } | null = null;
+
+  if (w < 320) {
+    preset = {
```
ilyabo: these presets could be defined as top-level consts

packages/vega/src/VegaLiteArrowChart.tsx
```diff
@@ -106,13 +155,16 @@ const VegaLiteArrowChartBase: React.FC<VegaLiteArrowChartProps> = ({
       padding: 10,
       background: 'transparent',
       ...parsed,
+      config: {
+        ...getResponsiveFontSizeConfig(dimensions.width),
```
ilyabo: I think this could be extracted into a useMemo too, which in combination with using consts for  presets should reduce unnecessary chart rerenders

packages/vega/src/VegaLiteArrowChart.tsx
```diff
@@ -106,13 +155,16 @@ const VegaLiteArrowChartBase: React.FC<VegaLiteArrowChartProps> = ({
...
+      },
       data: data,
       // Override the following props to ensure the chart is responsive
       width: 'container',
       height: 'container',
       autosize: {contains: 'padding'},
     } as VisualizationSpec;
-  }, [spec, data]);
+  }, [spec, data, dimensions.width]);
```
ilyabo: just depend on `config` here instead of `dimensions.width`, and config comes from another memo

packages/vega/src/VegaLiteArrowChart.tsx
```diff
@@ -106,13 +154,16 @@ const VegaLiteArrowChartBase: React.FC<VegaLiteArrowChartProps> = ({
       padding: 10,
       background: 'transparent',
       ...parsed,
+      config: {
+        ...getResponsiveFontSizeConfig(dimensions.width),
+      },
       data: data,
```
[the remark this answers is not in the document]
ilyabo: nice catch!

packages/vega/src/VegaLiteArrowChart.tsx
bdjulbic: Fix
Moved to top level

packages/vega/src/VegaLiteArrowChart.tsx
bdjulbic: Fix
Extracted to memo

packages/vega/src/VegaLiteArrowChart.tsx
bdjulbic: Fixed
Config comes from spec, 
so I just added responsiveFontConfig to the dependency array

packages/vega/src/VegaLiteArrowChart.tsx
```diff
@@ -106,13 +171,17 @@ const VegaLiteArrowChartBase: React.FC<VegaLiteArrowChartProps> = ({
       padding: 10,
       background: 'transparent',
       ...parsed,
+      config: {
+        ...((parsed as any)?.config ?? {}),
+        ...responsiveFontConfig,
+      },
```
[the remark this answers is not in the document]
bdjulbic: Is it ok If I use deepmerge utility for this?
`import deepmerge from 'deepmerge';`

packages/vega/src/VegaLiteArrowChart.tsx
bdjulbic: font sizing is removed now

PR #475, opened 2026-03-17
fix: update branch

bdjulbic:


packages/ui/src/components/sonner.tsx
```diff
@@ -19,18 +12,37 @@ const Toaster = ({...props}: ToasterProps) => {
       richColors
       theme={theme as ToasterProps['theme']}
       className="toaster group"
+      closeButton
+      swipeDirections={['bottom']}
```
bdjulbic, line 16: this locks the card swiping to only bottom direction
close button shows the close button on top left corner

PR #477, opened 2026-03-17
feat: implement internal resource filtering in createDbSchemaTrees

dmitriy-kostianetskiy:
Issue: https://github.com/sqlrooms/sqlrooms/issues/437

This pull request introduces comprehensive filtering logic to hide internal SQLRooms resources from the database schema tree and adds thorough tests to ensure this functionality. The main focus is to ensure that any databases or schemas prefixed with `__sqlrooms_` are excluded from the UI, while still allowing user-created resources with other `__` prefixes. The changes also include a basic test for tree structure creation.

Filtering internal resources:

* Added the `isInternalSqlRoomsResource` function in `schemaTree.ts` to identify internal SQLRooms resources by their name prefix (`__sqlrooms_`).
* Updated `createDbSchemaTrees` in `schemaTree.ts` to skip any databases or schemas identified as internal SQLRooms resources, ensuring they are not included in the schema tree.

Testing improvements:

* Added a suite of tests in `schemaTree.test.ts` to verify filtering of internal SQLRooms databases and schemas, correct handling of user databases with `__` prefix, mixed resource scenarios, and default database behavior.
* Included a basic functionality test in `schemaTree.test.ts` to confirm that the schema tree is correctly constructed for standard cases.

## Summary by CodeRabbit

* **New Features**
  * Added optional filtering capability for database schema loading

* **Tests**
  * Added validation tests for schema tree structure

* **Chores**
  * Removed unused code and updated repository configuration

### DIFF
--- a/packages/duckdb/src/DuckDbSlice.ts
+++ b/packages/duckdb/src/DuckDbSlice.ts
@@ -11,8 +11,12 @@ import {
   QualifiedTableName,
   QueryHandle,
   separateLastStatement,
-  TableColumn,
 } from '@sqlrooms/duckdb-core';
+import {
+  loadTableSchemas,
+  LoadTableSchemasFilter,
+  LoadTableSchemasFilterFunction,
+} from './loadTableSchemas';
 import {
   BaseRoomStoreState,
   createSlice,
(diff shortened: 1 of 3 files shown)


packages/duckdb/src/DuckDbSlice.ts
```diff
@@ -487,6 +488,17 @@ export function createDuckDbSlice({
             filter?: SchemaAndDatabase & {table?: string},
```
ilyabo: looks like Copilot didn't follow the instructions carefully 
> What I would suggest to do is adding an optional loadTableSchemasFilter?: (table: QualifiedTableName) => boolean prop to CreateDuckDbSliceProps and applying it in loadTableSchemas. Then we can pass the __sqlrooms filter in the CLI app.
Or maybe the default can filter all __sqlrooms* schemas and databases out.

dmitriy-kostianetskiy: I guess the idea was to filter out all `__sqlrooms*` schemas by default, which this change effectively does at the sql level.

do you want me to pass it via filter parameter? If yes I have to two points to clarify:
1. Current implementation is very effective sql-level **include** filter. We can add a function like you described, however it won't be part of SQL query anymore, but predicate for javascript `.filter`. If you think this is fine, I can refactor it this way.
2. More conceptually, why would we need to that in the first place? What are the use cases for that sort of flexible filtering?

PR #494, opened 2026-03-24
fix: adjust padding in notebook examples

dmitriy-kostianetskiy:
## Summary by CodeRabbit

* **Style**
  * Updated layout tile styling in canvas examples to use default presentation
  * Modified file dropzone height in the notebook's data sources panel


ilyabo (approved):
we could do the same in examples/canvas and examples/canvas-sync

PR #503, opened 2026-03-30
feat: inline sql query

bdjulbic:


packages/ai/src/tools/query/QueryToolResult.tsx
```diff
@@ -12,6 +12,8 @@ export type QueryToolRendererOptions = {
   showSql?: boolean;
   /** Optional custom value formatter for binary/geometry data */
   formatValue?: ArrowDataTableValueFormatter;
+  /** Optional callback to create a new SQL tab with the query */
+  onCreateSqlTab?: (query: string) => void;
```
ilyabo, line 16: I would suggest to name it `onSaveQuery()` everywhere

packages/data-table/src/DataTableModal.tsx
```diff
@@ -59,22 +63,50 @@ export type DataTableModalProps = {
...
+                  size="sm"
+                  variant="outline"
+                  className="h-7 gap-1"
+                  onClick={() => {
+                    onCreateSqlTab(props.query!);
+                    tableModal.onClose();
+                  }}
+                >
+                  <SquareTerminal className="h-3 w-3" />
+                  <span className="text-xs">Open in SQL Editor</span>
```
ilyabo, line 98:
```suggestion
                  <span className="text-xs">Save query</span>
```

PR #497, opened 2026-03-24
feat: Upgrade AI SDK to v6

mihajlo-stojkovic:
## Summary

Upgrades AI SDK from v5 to v6 (`ai@6.0.132`) across all packages and examples.

The core change is replacing `streamText` with `ToolLoopAgent` — tool execution now happens server-side inside an agent loop instead of being orchestrated client-side via `onChatToolCall`. This removes ~120 lines of dead plumbing (`waitForToolOutput`, `pendingToolCallResolvers`, `onChatToolCall`) and simplifies the transport layer significantly.

Sub-agent composition follows the same pattern: `processAgentStream` is replaced by `streamSubAgent`, which wraps a `ToolLoopAgent` + `createAgentUIStream`.

The other major addition is native support for the v6 `needsApproval` approval flow. Tools with `needsApproval: true` now pause the agent loop, stream an `approval-requested` state to the client, and resume when the user confirms or denies. This replaces our custom approval wiring.

**Note**: The needsApproval mechanism in the Vercel AI SDK works by pausing the chat-level loop and emitting an approval-requested state in the UI message parts. But createAgentUIStream / ToolLoopAgent runs its own internal loop, and it doesn't know how to pause and wait for user approval.

> The solution to support needsApproval for nested agent is:
> 
> Add sub-agent approval infrastructure to AgentUtils.ts: Extend AgentStreamStore and streamSubAgent to detect tool-approval-request chunks, surface them via the store, wait for user response, and restart the stream with the approval response.
> 
> Add approval state/methods to AiSlice.ts: A pendingSubAgentApprovals map and methods to request/resolve them.
> 
> Update ToolPartRenderer.tsx: Make AgentToolCallEntry render the approval UI for sub-agent tools with pending approvals.
> 
> Update BookHotelApprovalRenderer.tsx: Make it work with the sub-agent approval mechanism.

Renames: `addToolResult` → `addToolOutput`, `processAgentStream` → `streamSubAgent`.

### Breaking changes

- `addToolResult` → `addToolOutput` with new signature: `{tool, toolCallId, output}`
- `processAgentStream` → `streamSubAgent(agent, prompt, abortSignal)`
- Tool renderers may receive new states: `approval-requested`, `approval-responded`, `output-denied`
- Remote transport routes must migrate to `ToolLoopAgent` + `createAgentUIStreamResponse`

See `docs/upgrade-guide.md` section 0.29.0 for full migration instructions with code examples.

## Summary by CodeRabbit

* **Documentation**
[...]

PR #508, opened 2026-03-31
feat: show elapsed time for tool calls and reasoning box in AI chat

lixun910:
Screenshot 2026-03-30 at 11 27 15 PM

## Summary by CodeRabbit

* **New Features**
  * Tool execution timings are captured, persisted with assistant messages, and injected on save/error so elapsed displays survive reloads/exports.
  * Live/frozen elapsed-time badges for individual tool calls, agent sub-steps, and grouped tool sequences.
  * Reasoning/group headers now show processing state, live subtitles, and support custom tool display name mappings; default label changed to "Processing".
  * Added compact duration formatting and a hook to compute elapsed-time strings.

### DIFF
--- a/packages/ai-core/src/components/ToolPartRenderer.tsx
+++ b/packages/ai-core/src/components/ToolPartRenderer.tsx
@@ -95,7 +95,8 @@ const AgentProgressRenderer: React.FC<{
   agentToolCalls: AgentToolCall[];
   finalOutput?: string;
   reasoning?: string;
-}> = ({toolCallId, agentToolCalls, finalOutput, reasoning}) => {
+  depth?: number;
+}> = ({toolCallId, agentToolCalls, finalOutput, reasoning, depth = 0}) => {
   const toolRenderers = useStoreWithAi((s) => s.ai.toolRenderers);
   const liveProgress = useStoreWithAi((s) => s.ai.agentProgress[toolCallId]);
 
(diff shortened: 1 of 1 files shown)


packages/ai-core/src/components/ReasoningBox.tsx
```diff
@@ -1,23 +1,68 @@
...
 };
 export const ReasoningBox: React.FC<ReasoningBoxProps> = ({
   children,
   className,
   title,
   defaultOpen = false,
+  isRunning = false,
+  toolCallIds,
 }) => {
   const displayTitle = title ?? 'Thought';
```
ilyabo: For the generic label I would suggest to use "Worked for 4m 18s" instead of just "Thought". 

Ideally, if we could ask AI to generate a short phrase explaining what's happening, like "Finding datasets", "Analyzing data", "Creating a map", that would be fantastic.

See also [this](https://chatgpt.com/share/69cb7c9d-60d8-8384-92ca-3dfe73a388e5).

packages/ai-core/src/AiSlice.ts
```diff
@@ -454,9 +488,11 @@ export function createAiSlice<TTools extends ToolSet = ToolSet>(
         },
 
         setConfig: (config: AiSliceConfig) => {
+          const restoredToolTimings = collectToolTimingsFromConfig(config);
           set((state) =>
             produce(state, (draft) => {
               draft.ai.config = config;
+              draft.ai.toolTimings = restoredToolTimings;
             }),
           );
```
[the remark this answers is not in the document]
lixun910: yes, a wholesale replacement would discard in-flight entries — the precondition for the bug (calling setConfig while tools are executing) doesn't occur in the actual codebase. This is a false positive.

packages/ai-core/src/components/ReasoningBox.tsx
lixun910: Sounds good! Can I address this in a new PR and land this PR for now? THanks!

lixun910:
Update the reasoning box wording:

https://github.com/user-attachments/assets/2cf28b5a-b7be-4159-a5b4-f89cb4988629

lixun910:
Since we are going to migrate to AI SDK V6 which has different implementation of agent (with its own loop executiono f sub-tools), we can implement elapsed time in https://github.com/sqlrooms/sqlrooms/pull/497 directly.

PR #530, opened 2026-04-11
feat: add Cesium 3D globe visualization package

collord:
Introduces @sqlrooms/cesium, a new package for 3D geospatial visualization built on CesiumJS and Resium, along with an example app demonstrating earthquake data visualization on a 3D globe.

Package features:
- CesiumSlice for state management (camera, clock, layers, scene)
- SQL-driven entity layers via useSqlToCesiumEntities hook
- Support for geojson, czml, sql-entities, and tileset layer types
- Time-animated entities with clock sync
- Depth testing against terrain, subsurface rendering, and camera height controls
- CesiumPanel, CesiumClock, CesiumToolbar UI components

Example app (examples/cesium-app):
- Earthquake explorer loading parquet data from HuggingFace
- Demonstrates SQL query to 3D entity rendering pipeline
- Uses vite-plugin-cesium for asset handling

## Summary by CodeRabbit

* **New Features**
  * New Cesium 3D globe package with interactive viewer, persistent camera, layer management, timeline playback with speed control, camera "zoom to fit", SQL-driven entity layers, viewer hooks, and utilities
  * UI components: embedded clock controls, toolbar, panels, and Cesium viewer wrapper
  * Example app: "Cesium Earthquake Explorer" with preset camera views, slab slicing, live histogram, SQL editor and rendering status overlays
* **Documentation**
  * Package and example READMEs with setup, usage, and troubleshooting (including env token example)
* **Chores**
  * Added project configs, TypeScript/Vite setup, package manifests, scripts, and .gitignore entries


collord:
Addressed CodeRabbit feedback across two commits:

**`a647a2d6`** — correctness fixes:
- Removed manual `initialize()` call in `Room.tsx` (StrictMode double-init; `RoomShell` handles initialization automatically)
- `createDefaultCesiumConfig` now returns `structuredClone(DEFAULT_CESIUM_CONFIG)` so callers can't mutate the shared default
- `useSqlToCesiumEntities` skips rows with non-finite lon/lat/alt or unparseable time values before passing them to `Cartesian3.fromDegrees` / `JulianDate.fromIso8601`
- `useClockSync` uses `useShallow` to select only store→Cesium inputs, excluding `currentTime`. Previously the tick handler wrote `currentTime` to the store, retriggering the effect that reapplied the full clock config to Cesium, creating a feedback loop.

**`81c9ae77`** — cleanups:
- Removed debug `console.log` in `CesiumEntityLayer`
- Fixed `examples/cesium-app/README.md`: earthquake filter is magnitude 5.0+ (not 4.0+), and speed slider is logarithmic from 1x up to 1 year/sec (not "0.1x to 100x")
- Fixed the stale "All angles in radians" doc comment on `CameraPosition` — longitude/latitude are degrees; only heading/pitch/roll are in radians

collord:
Addressed the second round of CodeRabbit feedback in `6fc6aeb6`:

- **useClockSync**: Restored store→Cesium `currentTime` propagation. A `lastTickTimeRef` tracks values written by the tick handler; a separate effect watches store `currentTime` and applies it to the viewer only when the value differs from the last tick write. This prevents the feedback loop while allowing external `setCurrentTime` calls (e.g. timeline scrub) to reach the viewer. The mutation is extracted into a module-level `applyCurrentTimeToViewer` function to satisfy React Compiler's immutability rule (same pattern as the existing `applyClockConfigToViewer`).

- **useSqlToCesiumEntities**: `stopJd` is now computed by scanning all rows for the maximum valid timestamp via `JulianDate.greaterThan`, instead of assuming the last row is latest. Also moved the `mapping` fallback inside the `useMemo` callback to fix the `react-hooks/exhaustive-deps` warning (previously it created a new object on every render, invalidating the memo).

collord:
Addressed the remaining CodeRabbit comment in `0e42c01e`:

- **useSqlToCesiumEntities**: `label`, `color`, and `size` fields in entity descriptors are now guarded against null/undefined/NaN before coercion. Previously `String(null)` would produce `"null"` and `Number(undefined)` would produce `NaN`, leaking invalid values into Cesium entities.

ilyabo:
Thanks for the contribution!

I am not yet quite sure it's worth adding a dedicated package at this point (could you convince me otherwise? :) ), but an example would definitely be a valuable addition. In [this PR](https://github.com/sqlrooms/sqlrooms/pull/534) I am trying to make the example a bit more interesting and a better showcase of what Cesium can do by using opensky flight data which can be nicely animated:
image

image

what do you think?

collord:
Significantly expanded the example based on feedback that the demo needed more visual impact. The Cesium example is now a **Wadati–Benioff Explorer** — a 3D globe cross-section tool for subduction zones. Four new commits:

**`2d467ee8`** — `feat(cesium-app): 3D earthquake explorer with subduction zone slicer`
- Six presets (Tonga, Japan, Chile, Indonesia, Cascadia, Hellenic) that fly the camera to an along-strike view and SQL-filter the catalog to a thin section perpendicular to the trench.
- Live mag-vs-depth histogram panel (recharts) tied to the active slice.

**`bdd0f550`** — `feat(cesium-app): translucent Wadati–Benioff slab surface under seismicity`
- Parametric slab-top mesh rendered beneath the seismicity, tuned per-preset to Slab2.0 dip profiles so each zone has its own character (Cascadia flat, Tonga plunging to ~700 km).

**`0e726317`** — `fix(cesium): clipping plane lifecycle + layer clock re-sync on data change`
- Package-level fixes: removed a double-destroy crash on preset switch (Cesium's `ClippingPlaneCollection.setOwner` already destroys the prior collection; our manual `.destroy()` was clobbering it), and made `CesiumEntityLayer` re-sync the clock whenever entities change (previously a stale `currentTime` from the initial dataset would hide every entity from a filtered reload).

**`bf963c5a`** — `feat(cesium-app): section cut, global seismicity, preset polish`
- Single-plane globe clip (section cut) + two-plane slab-mesh clip (thin ribbon within ±halfWidth of the section line).
- SQL data-point filter uses 3× the slab width so the point cloud has lateral context.
- `flyToBoundingSphere` with a subsurface target so each preset auto-frames the full dipping zone.
- Dark scene controller (no stars/sun/atmosphere) for focus.
- `depthTestAgainstTerrain: true` — subsurface events hidden until a preset cuts the globe open.
- Swapped the bundled NA-only sample CSV for a global parquet (M5+ worldwide since 2013, ~17k events, ~850 KB) hosted on HuggingFace, matching the pattern used by `canvas`/`kepler`/`minimal`/`multi-room`. Temporarily at `huggingface.co/datasets/collord/globalearthquakes` — happy to move to `sqlrooms/*` whenever you tell me where.
- `scripts/fetch-earthquakes.py` regenerates the parquet from the USGS FDSN feed.
- `RenderingStatus` overlay covers the gap between parquet-download-done and entities-on-globe.

collord:
I updated my example to try to showcase why 3D matters for a general audience, still earthquakes though. I should be explicit that I'm personally interested in problems that involve the subsurface. Subsurface/3D spatial information systems is a nontrivial market that is addressed in very narrow domain-specific tools with typically very high per-seat costs. Examples are: support for tunnelling in the civil engineering domain (visualizing/modeling geotechnical data); visualization and modeling of subsurface environmental chemistry data in e.g. delineating a plume of a pollutant in groundwater; visualization and modeling of subsurface whole rock analytical or multispectral imaging data in a minerals exploration context; temperature in geothermal exploration; geophysics assets in all of the above, etc. etc. 

I am excited to have the real foundation for a frontend-heavy data-intensive experiments that duckdb-wasm and sqlrooms brings together.

Thank you for your consideration. 

> Thanks for the contribution!
> 
> I am not yet quite sure it's worth adding a dedicated package at this point (could you convince me otherwise? :) ), but an example would definitely be a valuable addition. 
> what do you think?

collord:
Addressed the third CodeRabbit round in `2154b4b1`:

- **`buildSliceWhereClause` cosLat accuracy**: swapped the preset-center `cos(lat)` for per-row `COS(RADIANS(latitude))`. The 1500 km along-section window spans ~13° of latitude, so a preset-center cosine could mis-scale edge points by ~20%. Remaining error is bounded by the flat-earth approximation itself (<1% for this extent).
- **Stale JSDocs**: the top-of-file comments in `store.ts` and `earthquake-slice.ts` still claimed `depthTestAgainstTerrain = false`. Updated to match the current behaviour (globe occludes subsurface events until a preset's section-cut clipping plane exposes them).
- **HistogramPanel selector stability**: gate `useSql` on a stable `Boolean(findTableByName(...))` rather than the table object, matching the pattern already used in `RenderingStatus` and the selector guidance in `CLAUDE.md`.
- **`preset.cameraHeight` cleanup**: removed the unused field. The camera range is now derived from slab depth via `flyToBoundingSphere`, so the field was no longer driving any behaviour.
- **`EARTHQUAKE_TABLE` dedupe**: exported the constant from `earthquake-slice.ts` and used it in `store.ts` (both the layer and data source `tableName`) and `HistogramPanel.tsx` instead of hardcoding `'earthquakes'`.

collord:
Fair point — on reflection, a dedicated Cesium package isn't worth the monorepo's long-term maintenance cost for what this is right now.

Pushed `7c284978` to fold `@sqlrooms/cesium` into a self-contained example. `examples/cesium-app` is renamed to `examples/cesium-quakes-standalone` and the package source lives at `examples/cesium-quakes-standalone/src/cesium/` (same structure — `cesium-config.ts`, `cesium-slice.ts`, `cesium-types.ts`, `utils.ts`, `components/`, `hooks/`), just no longer published as a workspace package.

No behaviour change — the example still renders the Wadati–Benioff globe with subsurface seismicity, section-cut clipping, and the slab surface mesh. Build + typecheck green across the monorepo (41 packages, 45 typechecks).

Re: the two examples — I don't have a strong opinion on which ships; happy for your opensky flight data branch (#534) to be the canonical Cesium example if you prefer. The quakes demo is there if you think the subsurface/clipping story adds something the flight demo doesn't, but no pressure.

PR #538, opened 2026-04-14
fix: WasmDuckDbConnector: worker not terminated on init failure

### LINKED ISSUE #531: WasmDuckDbConnector: worker not terminated on init failure, orphaning OPFS handles
## Bug

When `initializeInternal()` fails (e.g., OPFS `DataCloneError`), the catch block nulls the `worker` reference without calling `worker.terminate()`. The worker keeps running and holds OPFS file handles, preventing any subsequent OPFS initialization attempts from succeeding.

## Location

`packages/duckdb/src/connectors/WasmDuckDbConnector.ts` (or the compiled `.js`):

```typescript
// initializeInternal
try {
    // ...
    worker = new window.Worker(workerUrl);     // Worker created
    db = new duckdb.AsyncDuckDB(logger, worker);
    await db.instantiate(mainModule, pthreadWorker);
    await db.open({ ...restConfig, path });     // <-- Can fail (e.g., OPFS error)
    conn = await db.connect();
} catch (err) {
    db = null;
    conn = null;
    worker = null;   // <-- Reference lost, but worker.terminate() never called!
    throw err;
}
```

## Impact

After a failed init:
- `destroyInternal()` is a no-op because `worker`/`db`/`conn` are all `null`
- The orphaned worker still holds OPFS `FileSystemSyncAccessHandle` locks
- Retry attempts create NEW workers that conflict with the orphan:
  ```
  Error: Access Handles cannot be created if there is another
  open Access Handle or Writable stream associated with the same file.
  ```
- Every retry makes it worse (more orphaned workers, more locked handles)

## Fix

Terminate the worker before nulling the reference:

```typescript
catch (err) {
    if (conn) { try { await conn.close(); } catch {} }
    if (db) { try { await db.terminate(); } catch {} }
    else if (worker) { worker.terminate(); }
    db = null;
    conn = null;
    worker = null;
    throw err;
}
```

`AsyncDuckDB.terminate()` calls `this._worker.terminate()` internally, so calling it on `db` handles both. The `else if (worker)` branch covers the case where the worker was created but `db` instantiation failed.

## Context

This compounds with [duckdb/duckdb-wasm#2096](https://github.com/duckdb/duckdb-wasm/issues/2096), where the COI worker's OPFS initialization fails with `DataCloneError`. That's an upstream DuckDB-WASM issue, but the orphaned worker makes recovery impossible even for transient failures.

Found in `@sqlrooms/duckdb` 0.28.0-rc.0, confirmed still present in 0.29.0-rc.1.

ilyabo:
fixes https://github.com/sqlrooms/sqlrooms/issues/531

## Summary by CodeRabbit

* **Bug Fixes**
  * Added explicit cleanup procedures during initialization failures, including connection closure and resource termination before re-throwing errors.

### DIFF
--- a/packages/duckdb/src/connectors/WasmDuckDbConnector.ts
+++ b/packages/duckdb/src/connectors/WasmDuckDbConnector.ts
@@ -99,6 +99,22 @@ export function createWasmDuckDbConnector(
 
         conn = await db.connect();
       } catch (err) {
+        if (conn) {
+          try {
+            await conn.close();
+          } catch {
+            /* best-effort */
+          }
+        }
+        if (db) {
+          try {
+            await db.terminate();
+          } catch {
+            /* best-effort */
+          }
+        } else if (worker) {
+          worker.terminate();
+        }
         db = null;
         conn = null;
         worker = null;


charlieforward9:
Thank you! Intend to release this soon? I see you haven't made a release in a few weeks.0

PR #541, opened 2026-04-16
feat: show all schemas in datasource panel even if empty

bdjulbic:
By default datasource panel shows only schemas with tables.
It will ignore showing schemas without tables

This PR changes that and will show all schemas even if empty,
except for reserved names

## Summary by CodeRabbit

* **New Features**
  * Empty schemas now appear in the schema tree; schema discovery now collects all schemas so empty ones are shown.
  * New default visibility filter is exposed plus a factory to compose custom table/schema hiding rules.

* **Bug Fixes**
  * Reserved/internal schemas and tables are excluded from the schema browser.
  * Tree placeholder appears only when truly empty; leaf node spacing aligned.
  * Refreshes update tables and schema tree together to avoid unnecessary UI churn.

* **Tests**
  * Added tests covering empty-schema visibility and hide rules.

Screenshot 2026-04-22 at 14 32 19


packages/kepler/src/KeplerSlice.ts
```diff
@@ -467,8 +467,8 @@ export function createKeplerSlice({
             },
           });
           updateForwardDispatch();
-          updateMapConfigs();
           await get().kepler.syncKeplerDatasets();
+          updateMapConfigs();
           requestMapStyle(config.currentMapId);
```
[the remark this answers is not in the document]
bdjulbic: this issue dissapered after merging latest main branch

packages/duckdb/src/DuckDbSlice.ts
```diff
@@ -34,17 +35,29 @@ import {createWasmDuckDbConnector} from './connectors/createDuckDbConnector';
 
 const DUCKDB_COMMAND_OWNER = '@sqlrooms/duckdb';
 const INTERNAL_SQLROOMS_PREFIX = '__sqlrooms_';
+const HIDDEN_SCHEMAS = ['fsq_rag', 'temp', 'fsq_spatial'];
```
ilyabo, line 38: This is client app-specific. We shouldn't mention anything about fsq in sqlrooms.

packages/duckdb/src/DuckDbSlice.ts
```diff
@@ -302,10 +315,15 @@ export type DuckDbSliceState = {
 export type CreateDuckDbSliceProps = {
   connector?: DuckDbConnector;
   /**
-   * Optional filter function to control which tables are included when loading schemas.
-   * By default, filters out tables/schemas/databases starting with '__sqlrooms_'.
+   * Optional filter function to control which tables/schemas/databases are visible in the data source panel.
+   *
+   * Default behavior (createDefaultLoadTableSchemasFilter) excludes:
+   * - Internal SQLRooms items (prefixed with '__sqlrooms_')
+   * - Reserved schemas: 'fsq_rag', 'temp', 'fsq_spatial'
```
ilyabo, line 322: We should remove all mentions of fsq, 'temp' as well, let's only keep '__sqlrooms'

packages/duckdb/src/DuckDbSlice.ts
ilyabo: Client apps can provide their own custom implementations of loadTableSchemasFilter. It's currently not possible, but if you cherry-pick [this commit](https://github.com/sqlrooms/sqlrooms/commit/bb05c7e3d599537b0498636c09ea392b3f32a600), then it will be possible.

packages/ui/src/components/tree.tsx
```diff
@@ -48,7 +48,12 @@ function TreeNode<T>(props: TreeNodeProps<T>): React.ReactElement | null {
   const hasChildren = Boolean(children?.length);
   const [isOpen, setIsOpen] = useState(Boolean(treeData.isInitialOpen));
   if (!hasChildren) {
-    return <div className="pl-4">{renderNode(treeData, isOpen)}</div>;
+    return (
+      <div className="flex w-full items-center space-x-1">
+        <div className="shrink-0" style={{width: '18px'}} />
```
bdjulbic, line 53: this width fixes the alignment  for empty schemas and schemas with tables (which have an dropdown arrow)

Issue #548, opened 2026-04-18
@sqlrooms/deck GeoArrow-First deck.gl Integration


ilyabo:
Plan built with the help of gpt-5.4:

Draft implementation: https://github.com/sqlrooms/sqlrooms/pull/549

---

## Summary
- Build `packages/deck` around a dataset registry from the start, so multi-layer and multi-dataset specs are first-class instead of an add-on.
- Keep `@deck.gl/json` as the canonical spec format, but add SQLRooms-owned layer metadata for dataset binding and geometry overrides.
- Optimize supported layers for GeoArrow-native rendering, with a modular WKB/WKT fallback path that can later be replaced by generic TS WKB -> GeoArrow work.

## Public API
- Main React API:
  - `DeckMap`
  - Props:
    - `spec: string | object`
    - `datasets?: Record<string, DeckDatasetInput>`
    - single-dataset sugar: exactly one of `sqlQuery`, `arrowTable`, or `queryResult`, internally normalized to `datasets.default`
    - `mapStyle?`, `deckProps?`, `mapProps?`, `className?`
- Dataset input type:
  - `DeckDatasetInput = { sqlQuery?: string; arrowTable?: arrow.Table; queryResult?: unknown; geometryColumn?: string; geometryEncodingHint?: 'geoarrow' | 'wkb' | 'wkt' }`
  - exactly one of `sqlQuery`, `arrowTable`, or `queryResult` is allowed per dataset entry
- Spec contract additions:
  - `sqlroomsData?: string` on each SQLRooms-managed layer
  - `sqlroomsGeometryColumn?: string` as a per-layer override of the dataset default
  - `sqlroomsGeometryEncodingHint?: 'geoarrow' | 'wkb' | 'wkt'` as an escape hatch
- Binding rules:
  - if `datasets` has one entry and a layer omits `sqlroomsData` and `data`, auto-bind that layer to the sole dataset
  - if `datasets` has multiple entries, every SQLRooms-managed layer must set `sqlroomsData`
  - if a layer provides explicit non-SQLRooms `data` and no `sqlroomsData`, leave `@deck.gl/json` behavior unchanged
  - each layer binds to exactly one dataset id in v1

## Implementation Changes
- Dataset registry runtime:
  - normalize all inputs into an internal registry keyed by dataset id
  - track per-dataset state: `idle | loading | ready | error`
  - prepare datasets independently so ready layers can render while others are still loading
- Geometry preparation:
[...]

PR #549, opened 2026-04-18
feat:  Deck.gl + geoarrow layers integration

ilyabo:
solves #548 
image

## Summary by CodeRabbit

* **New Features**
  * New color scale package and legend UI for configurable color legends.
  * New deck package and a DeckJsonMap component for spec-driven map rendering and layer binding.
  * MosaicColorLegend component for Mosaic charts.

* **Improvements**
  * Examples updated to use the new map and color-scale flow with theme-aware styles and integrated legends.
  * Simplified Mosaic hook usage (no explicit generic type).

* **Documentation**
  * Added table interop contract and README/API documentation requirement.


packages/mosaic/src/profiler/useMosaicProfiler.ts
dzole0311, line 1: I think we could speed up the cross filtering between the charts and the table by building MosaicProfiler to instantiate vg.table() via the mosaic package. Similar for the summaries in the table, we could use the vg.chart().

dzole0311: Something like this: https://github.com/sqlrooms/sqlrooms/pull/550

PR #550, opened 2026-04-19
feat: integrate vg.table() for the mosaic deck demo

dzole0311:
Related: https://github.com/sqlrooms/sqlrooms/pull/549

Opening this to showcase a vg.table()-based deck.gl + Mosaic profiler flow in the deckgl-mosaic example.


ilyabo:
Thanks Gjore! It feels faster indeed, especially the table, but the setup is quite involved, and customizations are less straightforward. So I would probably stick to the React components at least for now. But we should keep in mind that there's such option. The infinite scrolling in the table is nice.

PR #564, opened 2026-04-23
feat: adding a table from non main schema

bdjulbic:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table naming/labels when adding tables to maps by collapsing redundant schema prefixes for clearer names.
  * Dataset synchronization now attempts to add any referenced dataset (not limited to a single schema) and logs individual failures without interrupting the sync.

* **Chores**
  * Removed an unused loader dependency from a package.


packages/kepler/src/KeplerSlice.ts
```diff
@@ -685,21 +695,23 @@ export function createKeplerSlice({
...
-
                 for (const dataId of referencedDataIds) {
-                  if (
-                    !keplerDatasets?.[dataId] &&
-                    availableTables.has(dataId)
-                  ) {
+                  if (keplerDatasets?.[dataId]) {
+                    continue;
+                  }
+                  if (!availableTableIds.has(dataId)) {
```
ilyabo: this is checking whether a referenced table is a view, let's skip this check completely, we should allow users to add views by default, or can assume that this is prevented by the ui restrictions in client apps

packages/duckdb-core/src/duckdb-utils.ts
```diff
@@ -37,6 +37,66 @@ export function makeQualifiedTableName({
...
+ * Each key represents a fully qualified table name.
+ *
+ * Previously, sqlrooms used only the table name as the key. This function maintains backward
+ * compatibility with that behavior.
+ *
+ * If a table is not fully qualified, two entries are created:
+ * - main.{tableName}
+ * - {tableName}
+ */
+export function buildKeplerUserTableIdMap(
```
ilyabo: let's remove this function, see below

packages/duckdb-core/src/duckdb-utils.ts
```diff
@@ -37,6 +37,66 @@ export function makeQualifiedTableName({
...
+ * Extracts the table name from a fully qualified SQL identifier.
+ * Removes database/schema prefixes and SQL quoting.
+ *
+ * Examples:
+ * - `"db"."schema"."table"` → `table`
+ * - `schema.table` → `table`
+ * - `"my""table"` → `my"table` (unescapes doubled quotes)
+ * - `table` → `table`
+ */
+export function unqualifySqlTableName(tableSql: string): string {
```
ilyabo: I think it would be good to move it next to [`makeQualifiedTableName()`](https://github.com/sqlrooms/sqlrooms/blob/main/packages/duckdb-core/src/duckdb-utils.ts#L23)

ilyabo: Also I think it would be good to add test cases for it to make sure it handles edge cases correctly, like, table names containing various quotes or dots

packages/kepler/src/KeplerSlice.ts
```diff
@@ -605,7 +609,10 @@ export function createKeplerSlice({
           if (fields && cols) {
             const datasets: AddDataToMapPayload['datasets'] = {
               data: {fields, cols, rows: [], arrowTable: arrowResult},
-              info: {label: tableName, id: tableName},
+              info: {
+                id: tableName,
+                label: unqualifySqlTableName(String(tableName)),
```
ilyabo: actually, I think that we should use fully qualified names here unless it's the default DB and schema, so maybe there's no need in `unqualifySqlTableName()`

packages/duckdb-core/src/duckdb-utils.ts
ilyabo: actually, maybe we don't need this function either 🤦  see below

PR #619, opened 2026-05-13
feat: dashboard empty state

dmitriy-kostianetskiy:
image


packages/mosaic/src/dashboard/MosaicDashboardEmptyState.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <BarChart3 className="mr-2 h-4 w-4" />
+            Chart
+          </Button>
+          <Button
+            variant="outline"
+            disabled={!canAddProfiler}
+            onClick={handleAddProfiler}
+          >
+            <TableProperties className="mr-2 h-4 w-4" />
+            Profiler
```
ilyabo, line 105: Now I agree with you, I think we should rename it to "Data table"

ilyabo (approved):

PR #661, opened 2026-05-25
feat: add overlaid integration mode to DeckJsonMap

### LINKED ISSUE #580: Allow controlling MapLibre/deck.gl integration mode in @sqlrooms/deck
In the deckgl example, the 3D Zurich buildings layer visually covers the MapLibre zoom controls and attribution/contribution bar. This makes the map controls hard to use when some of the layers are enabled.

This likely comes from the current integration mode in `@sqlrooms/deck` where `DeckJsonMap` renders `DeckGL` as the root and nests the MapLibre `Map` (which matches deck.gl's reverse-controlled mode and does not support MapLibre controls/plugins). For MapLibre controls, deck.gl recommends using `MapboxOverlay` in overlaid or interleaved mode.

Docs: https://deck.gl/docs/developer-guide/base-maps/using-with-maplibre

### Possible approach

Add a mode option to `DeckJsonMap`, for example:

- `reverseControlled` / current behavior
- `overlaid` using `MapboxOverlay({ interleaved: false })`
- `interleaved` using `MapboxOverlay({ interleaved: true })`

The deckgl buildings example probably wants `overlaid` or `interleaved`.

igorDykhta:
## Summary

- Replace the `DeckGL` wrapper with `MapboxOverlay` integration so MapLibre controls and attribution render above deck layers
- Add `interleaved` prop to `DeckJsonMap` (defaults to `false`)
- Add `@deck.gl/mapbox` dependency
- Remove `useDeckLayersReadyRedraw` (no longer needed with overlay-based rendering)

The two modes are:

**Overlaid** (default, `interleaved=false`) — MapLibre is the root; deck.gl renders in a separate overlay canvas on top via `MapboxOverlay`. MapLibre controls and attribution stay accessible. Deck layers always render above all basemap layers.

**Interleaved** (`interleaved=true`) — MapLibre is the root; deck.gl layers are inserted into MapLibre's layer stack sharing the same WebGL2 context. This lets you render deck layers between basemap layers (e.g., put points under map labels using `beforeId`). Requires WebGL2 (MapLibre GL v3+).

For dashboards, overlaid (the default) is the right choice. Interleaved would only matter if you wanted deck layers to appear beneath map labels or road names, which is a niche use case.

Fixes #580

Screenshot 2026-05-25 at 3 12 54 AM
Screenshot 2026-05-25 at 3 23 04 AM


ilyabo:
Actually, maybe we don't need `reverse-controlled` at all esp. if it's legacy mode, and could just use `interleaved: boolean` prop which DeckOverlayControl already has, then we don't need  to introduce `integrationMode` and a separate render path to support. What do you think?

ilyabo (approved):

PR #671, opened 2026-06-01
feat: Adding a datasource from a non main schema using fully qualified table name 

bdjulbic:
Use fully qualified table name when adding data to the map.
This allows adding data to a map from a non-main schema.


packages/cells/src/helpers.ts
```diff
@@ -101,59 +101,7 @@ export function normalizeCellsConfigStructure(
...
- *
- * The parser is quote-aware: dots inside double-quoted identifiers are treated
- * as part of the identifier rather than as qualification separators.
- *
- * Examples:
- * - `schema.table` -> `table`
- * - `db.schema.table` -> `table`
- * - `schema."my.funny.table"` -> `my.funny.table`
- */
-export function getUnqualifiedSqlIdentifier(
```
bdjulbic, line 115: this was moved to packages/duckdb-core/src/duckdb-utils.ts

ilyabo (approved):

PR #674, opened 2026-06-02
fix: save map layers

bdjulbic:
## Release Notes

* **Bug Fixes**
  * Enhanced map configuration auto-save logic to prevent saving incomplete states during map restoration
  * Added safeguards to protect existing configurations from being unintentionally overwritten during application startup and data synchronization

### DIFF
--- a/packages/kepler/src/KeplerSlice.ts
+++ b/packages/kepler/src/KeplerSlice.ts
@@ -303,6 +303,12 @@ export function createKeplerSlice({
   // active run to loop once more with a fresh snapshot after it finishes, so
   // late mutations are never silently dropped.
   let pendingKeplerSync = false;
+  // While restoring saved configs (setConfig -> updateMapConfigs), the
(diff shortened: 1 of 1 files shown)


packages/kepler/src/KeplerSlice.ts
```diff
@@ -872,17 +895,45 @@ export function createKeplerSlice({
...
+                  const newDatasets = Object.keys(
+                    state.kepler.map[mapId].visState?.datasets ?? {},
+                  ).length;
+                  if (
+                    existingLayers > 0 &&
+                    newLayers === 0 &&
+                    newDatasets === 0
+                  ) {
+                    return;
+                  }
```
bdjulbic, line 935: I am not familiar with react-palm. 
Is someone in the team familiar with this?

ilyabo: @igorDykhta might be able to help with this

PR #689, opened 2026-06-04
feat: support brush selection in vega-lite chart

lixun910:
## Summary

Add brush selection functionality to Vega-Lite chart components and integrate with map context.

Cherry-picked from commit `ef42ab5dc964c084f6791e62e052fbc4fca971e4`.

1. AI creates a chart with e.g.:
     sqlQuery:  "SELECT city, AVG(price) AS avg_price FROM sales GROUP BY city"
     vegaLiteSpec: bar chart with x=city (nominal), y=avg_price (quantitative)

```
│ city     │ avg_price │
│ NYC      │ 35        │
│ LA       │ 22        │
│ SF       │ 45        │
│ Chicago  │ 18        │
```

**Note: if there are more than 1 table involved in vega-chart, then brush disabled. We can instruct AI to always save the complex query to one table for charting.**

2. VegaLiteArrowChart detects:
     x is nominal, y is quantitative
     → injects interval brush on y-axis only

3. User drags a brush on the y-axis selecting avg_price range [20, 40]
     → Vega emits signal: { avg_price: [20, 40] }

**Note: for e.g. pie chart without x, y encoding, it will fall back to `point selection` and normalizePointSignal() will be used to return all matching source row indices**

4. KeplerVegaChartToolResult receives the signal:
     First time → calls `computeBrushMapping(sqlQuery)`
       - Extracts table: "sales"
       - Extracts non-aggregate aliases: { city → "city" }
         (avg_price is skipped — it's an aggregate)
       - Queries: SELECT row_number()-1 AS __row_idx, city FROM sales
       - Builds mapping: { city: { "NYC": [0,3,7], "LA": [1,4], ... } }
       
**Note: if there is no matching map layer: A dimmed/muted mouse-pointer icon with tooltip "No map layer linked". No brush interaction is possible on the chart.**

5. `resolveBrushToRowIndices`(mapping, { avg_price: [20, 40] }, chartData)

  Step A — avg_price is NOT in the mapping → aggregate bridge kicks in
  Step B — Scan the chart's Arrow table for rows where avg_price ∈ [20, 40]:
```
       NYC     → 35 ✓
       LA      → 22 ✓
       SF      → 45 ✗
       Chicago → 18 ✗
```
  Step C — Collect the group-by keys (fields that ARE in the mapping):
```
       Matching groups: ["NYC", "LA"]
```
  Step D — Look up group keys in the existing mapping:
```
       city.NYC     → source rows [0, 3, 7]
       city.LA      → source rows [1, 4]
```
  Step E — Union: [0, 1, 3, 4, 7]

6. highlightRows() sets dataset.filteredIndex = [0, 1, 3, 4, 7]
     → Kepler re-renders showing only the matching points on the map

7. User clears the brush (clicks background)
     → Vega emits signal: {}
     → resolveBrushToRowIndices returns []
     → highlightRows resets filteredIndex to allIndexes
     → Map shows all points again

## Changes

- Add brush selection support to `VegaLiteArrowChart` and `VegaChartDisplay` components
- Add `brushMappingUtils` for mapping brush ranges to data fields
- Add `KeplerVegaChartToolResult` component for Kepler-Vega integration
- Expose `onBrushSelection` and `brushAvailable` props in `VegaChartToolResult`
- Update exports in kepler and vega packages

## Release Notes

* **New Features**
  * Added interactive chart brush selection with visual indicator showing map integration availability
  * Chart selections now automatically synchronize and highlight corresponding dataset rows in Kepler.gl map layers

* **Improvements**
  * Enhanced context message formatting in summarization workflow

### DIFF
--- a/packages/ai-core/src/components/ContextUsageIndicator.tsx
+++ b/packages/ai-core/src/components/ContextUsageIndicator.tsx
@@ -151,7 +151,7 @@ export const ContextUsageIndicator: React.FC<ContextUsageIndicatorProps> = ({
       const newSessionId = state.ai.config.currentSessionId;
       if (!newSessionId) return;
 
-      const contextMessage = `Please use the following context from a previous session and only respond "Got it" to this message:\n\n${summary}`;
+      const contextMessage = `Please use the following context from a previous session:\n\n${summary}\n\nONLY respond "Got it" to this message`;
       state.ai.setPrompt(newSessionId, contextMessage);
 
       // Wait for SessionChatProvider to mount and register sendMessage for the new session
--- a/packages/kepler/package.json
+++ b/packages/kepler/package.json
@@ -41,6 +41,7 @@
     "@sqlrooms/room-shell": "workspace:*",
     "@sqlrooms/s3-browser": "workspace:*",
     "@sqlrooms/ui": "workspace:*",
+    "@sqlrooms/vega": "workspace:*",
     "apache-arrow": "17.0.0",
     "immer": "^11.0.1",
     "lucide-react": "^0.556.0",
--- a/packages/kepler/src/components/KeplerVegaChartToolResult.tsx
+++ b/packages/kepler/src/components/KeplerVegaChartToolResult.tsx
@@ -10,7 +10,7 @@ import {
   extractTableNameFromSql,
   resolveBrushToRowIndices,
 } from '@sqlrooms/vega';
-import {ReactNode, useCallback, useEffect, useMemo, useRef} from 'react';
+import {ReactNode, useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {useStoreWithKepler} from '../KeplerSlice';
 import type {DbSliceState} from '@sqlrooms/room-shell';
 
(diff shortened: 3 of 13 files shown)


ilyabo:
@lixun910 this is looking pretty cool, only I am bit worried that it is adding confusion because we are currently working on mosaic-based charts and cross-filtering dashboards.

I asked gpt-5.5 about it and also to review the PR:

-----
Yes, I think it risks adding confusion right now, especially in the shadow of Mosaic-based charts and dashboard cross-filtering.

The feature itself is desirable: chart selection driving map/dashboard state is absolutely in SQLRooms’ direction. But this PR implements a fairly bespoke Vega-to-Kepler bridge: parse chart SQL, infer one source table, rewrite SQL into a row-index mapping query, mutate Kepler dataset filtering, and expose that through a specialized `KeplerVegaChartToolResult`. That is a lot of local machinery for one integration path.

My concern is not “brush selection is bad.” It is that this creates a second cross-filtering model before the Mosaic/dashboard story has settled. If Mosaic charts and dashboards are where cross-filtering should live, then this PR may become an attractive but awkward parallel path: Vega chart brush signals go through SQL AST mapping into Kepler row indices, while Mosaic/dashboard filters likely want shared, inspectable filter state and composable query predicates.

So my read:

- **Net positive as a prototype:** yes, it explores a useful interaction.
- **Net positive to land as-is:** no, I would not.
- **Accidental complexity risk:** high.
- **Implementation soundness:** not yet, because the aggregate example and multi-table guard do not match the code.
- **Strategic fit:** shaky unless it is reframed around a common selection/filter primitive.

The better direction, I think, is to extract the core idea into a small shared concept: “chart emits selection over fields,” then let dashboards/maps/Mosaic consume that through common filter state or commands. Vega and Mosaic can both be producers. Kepler can be one consumer. That keeps the behavior composable instead of baking cross-filtering into a Vega-Kepler one-off.

-----
As written, I would not land this PR. The direction is net-positive and the base Vega behavior is mostly opt-in: `VegaChartToolResult` only injects brush params when `onBrushSelection` is provided. It is not wholesale for all charts.

But the Kepler wrapper makes the new behavior effectively automatic for charts whose SQL table name matches a map layer, and the implementation is not sound enough yet. The main aggregate use case in the PR description is missing in code, and joins/multi-table queries are not disabled despite the PR saying they are.

**Findings**

1. **Blocker: the headline aggregate brush case does not work.** In `packages/vega/src/brushMappingUtils.ts:306`, `resolveBrushToRowIndices()` only resolves fields that already exist in `mapping`. But `computeBrushMapping()` intentionally skips aggregate fields at `packages/vega/src/brushMappingUtils.ts:233`, so a brush like `{ avg_price: [20, 40] }` is ignored and returns `[]`. `KeplerVegaChartToolResult` calls it with only `mapping` and `ranges` at `packages/kepler/src/components/KeplerVegaChartToolResult.tsx:173`, so the “aggregate bridge” described in the PR body is not implemented.

2. **Blocker: multi-table/join SQL is not actually disabled.** `findBaseTableName()` just returns the leftmost table at `packages/vega/src/brushMappingUtils.ts:86`, and `hasMatchingLayer` marks the chart brushable if that table has a map layer at `packages/kepler/src/components/KeplerVegaChartToolResult.tsx:92`. But `computeBrushMapping()` then builds row indices from the rewritten full query at `packages/vega/src/brushMappingUtils.ts:256`, so joins can produce row numbers that do not correspond to the Kepler dataset’s source rows. That can highlight wrong map rows.

3. **Correctness risk: chart SQL filters are stripped.** The mapping query explicitly drops `WHERE`, `HAVING`, `QUALIFY`, and grouping at `packages/vega/src/brushMappingUtils.ts:263`. That means brushing a chart built from filtered data can highlight rows that were not part of the chart. Maybe that is intended, but it is surprising and should be made explicit or fixed.

4. **Accidental complexity: lockfile churn looks unrelated.** `pnpm-lock.yaml` changes much more than adding `@sqlrooms/vega` to `@sqlrooms/kepler`, including peer resolution and optional platform metadata changes. I would regenerate it with the repo’s expected pnpm version and keep only necessary dependency changes.

-----

PR #694, opened 2026-06-05
feat: inline vegalite chart editing

lixun910:
WYSIWYG editing on top of Vega-Lite's rendered SVG
  - edit title
  - drag-n-move text labels
  - delete text labels

other:
- update style when rendering agent reasoning text
  
Screenshot 2026-06-05 at 11 20 02 AM

## Summary by CodeRabbit

* **New Features**
  * Added interactive chart editing with WYSIWYG title/subtitle editing, data-label repositioning, and deletion capabilities.
  * Introduced an optional interactive chart tool renderer with automatic persistence of edits to the AI session.
  * Enhanced chart export functionality for improved SVG and PNG quality with better style handling.

* **Documentation**
  * Added documentation for the new interactive chart tool renderer with usage examples.

* **Style**
  * Updated visual styling for reasoning text display.


lixun910:
@ilyabo A candidate feature for vega-lite chart tool

ilyabo:
@codex review

packages/ai-core/src/index.ts
```diff
@@ -103,3 +103,12 @@ export {
   HoistedRenderersProvider,
   useHoistedRenderers,
 } from './components/HoistedRenderersContext';
+export {
+  findChatSearchMatches,
```
ilyabo, line 131: looks like this is from another PR

packages/ai-core/README.md
```diff
@@ -102,6 +102,30 @@ message and composer components stay under the same `Chat` compound API.
 </Chat.LocalAgentRoot>
 ```
 
+## Chat search
+
+`Chat` renders a `ChatSearchProvider` and exposes `Chat.Search`, an in-conversation
```
ilyabo, line 120: looks like this is from another PR

packages/ai/src/index.ts
```diff
@@ -144,6 +144,12 @@ export type {
 } from '@sqlrooms/ai-core';
 export {Chat} from '@sqlrooms/ai-core';
 export {ContextSelector, CHAT_CONTEXT_SELECTOR_SLOT} from '@sqlrooms/ai-core';
+export {
```
ilyabo, line 167: looks like this is from another PR

ilyabo:
@lixun910  This looks cool, but wouldn't it be more valuable, instead of making the charts UI more complex, to add the functionality within individual charts to ask AI to make adjustments in them? e.g. something like this (Databricks):

image
image


also we are trying to invest more in mosaic charts now, than vega, because they have more powerful cross-filtering capabilities.


AI review (with my input):

----

Here’s the combined version with that added:

---

I’m supportive of the goal here: letting users polish generated charts is meaningful, especially for presentation/reporting workflows. But I think this PR may be pushing too much complexity into the Vega chart UI itself, and I’m not sure this is where we should invest deeply right now.

My bigger product concern is that the most valuable version of this might be “ask AI to adjust this chart” rather than a growing WYSIWYG editing toolbar. For SQLRooms, that seems more aligned with the architecture: the AI can operate on the same chart primitives users already have, propose spec/query changes, preview them, and persist normal chart state. That would work for both AI-generated and manually-created charts.

There’s also a strategic consideration: we’re increasingly investing in Mosaic charts rather than Vega, because Mosaic gives us stronger cross-filtering and interactive analytical capabilities. So I’d be cautious about building a complex Vega-specific editing surface unless we’re confident it will become a reusable pattern across chart systems.

Right now the PR feels mostly aimed at AI-generated chat charts. `InteractiveVegaChartToolResult` persists edits into AI session `uiMessages`, so the first-class persistence path is chat-specific. There is a reusable lower-level `VegaInteractiveEdit` component, which is good, but the documented and packaged flow is still centered on AI tool results. For non-AI charts, I’d want the main abstraction to be caller-owned chart state: current spec/query in, `onSpecChange`/`onSqlChange` out, with no dependency on AI session internals.

A few implementation concerns before merge:

- Dragged labels are persisted as absolute SVG `translate(x, y)` positions, not true offsets from Vega’s recomputed label position. That is fragile across resize, data changes, theme/layout changes, and different render contexts. It is probably okay for static post-generation polish, but not robust general chart editing.
- Pending edits can be lost if the chart unmounts before the debounce fires. The cleanup clears the timeout without flushing the session update.
- The chat-search exports/docs look unrelated to this PR and should probably be split out.
- The lockfile churn is much broader than adding `immer`; it rewrites unrelated peer-resolution metadata and should be regenerated/cleaned.

My recommendation would be to narrow this PR: keep any generally useful export improvements and maybe the reusable primitive if we’re confident in it, but avoid making the AI chat renderer the main abstraction. Longer term, I’d rather see a chart-local AI action like “Adjust chart…” that receives the current chart spec/config, SQL/query, schema/sample, and user instruction, then returns a previewable change. Ideally that pattern should work for Mosaic first, and Vega only where it remains useful.

----

Findings:

[P1] Dragged labels persist as absolute SVG positions, not offsets. In [VegaInteractiveEdit.tsx](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/packages/vega/src/VegaInteractiveEdit.tsx#L585), finalDx/finalDy are the final translate(...) coordinates and later reapplied verbatim at [line 776](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/packages/vega/src/VegaInteractiveEdit.tsx#L776). That means resize, data changes, or different chart layout can pin a label to an old absolute pixel location instead of new Vega position + user delta. This is probably acceptable only for static post-generation polish, not durable/non-AI chart editing.

[P2] Pending AI chart edits can be lost on unmount/session switch. [InteractiveVegaChartToolResult.tsx](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/packages/vega/src/InteractiveVegaChartToolResult.tsx#L141) clears the debounce timer without flushing it, while the only session write happens later at [line 185](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/packages/vega/src/InteractiveVegaChartToolResult.tsx#L185). A user can edit, switch sessions/reload within 500ms, see the local cache mask it briefly, then lose the edit. Flush on cleanup or persist synchronously when unmounting.

[P2] The PR has unrelated public API/docs changes. The chat search exports/docs in [packages/ai-core/README.md](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/packages/ai-core/README.md#L105), [packages/ai-core/src/index.ts](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/packages/ai-core/src/index.ts#L107), and [packages/ai/src/index.ts](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/packages/ai/src/index.ts#L148) look like another PR. I’d split them out.

[P2] Lockfile churn is far broader than the dependency addition. Adding immer should not rewrite unrelated peer resolutions like @loaders.gl/parquet/ws at [pnpm-lock.yaml#L20834](https://github.com/sqlrooms/sqlrooms/blob/425b47bca4816acefbd230807bd49c17fdfc5a1c/pnpm-lock.yaml#L20834). Regenerate with the repo’s pinned pnpm and keep only the real @sqlrooms/vega/immer changes.

PR #700, opened 2026-06-08
feat: improvements for geoarrow deck layers in dashboard maps

igorDykhta:
### New layer implementations
- **H3HexagonLayer** — Custom GeoArrow-compatible wrapper that converts Arrow data to row objects with proper BigInt→hex string conversion, working around incompatibilities between `@geoarrow/deck.gl-layers@0.3.2` and `@deck.gl/geo-layers@9.3.x`
- **TripsLayer** — Custom GeoArrow-compatible wrapper that extracts binary data from Arrow vectors and passes it to the native TripsLayer, with proper per-chunk iteration and Float32 timestamp conversion
- **ColumnLayer** — Wrapper that patches a deck.gl wireframe index buffer bug (#10021) affecting fill rendering with binary data
- **Waypoint aggregation** — Auto-aggregates per-waypoint rows into LineString vectors for PathLayer/TripsLayer by detecting group-by ID, order, and timestamp columns via naming patterns
### Fit-to-bounds refactor
- Extracted fit-to-bounds logic into `useDeckMapFitToBounds` hook with support for H3, geometry, and lat/lon column types
- Added `jumpTo` imperative handle on `DeckJsonMap` via `forwardRef`
- Bounds queries now handle WKB-encoded geometry columns and H3 cell centroids
### Map configuration UI
- Added extrusion controls (toggle + elevation column picker) for H3 and polygon layers
- Added trips timeline control with play/pause, scrubber, and collapse
- Added missing-column warning icon overlay that opens settings on click
- Highlighted missing columns in `ColumnSelector` with red border and "missing" label
- Made `ColorScaleLegend` collapsible
### AI instructions & tooling
- Improved instructions for H3 functions (`h3_cell_to_lat`/`h3_cell_to_lng` instead of `h3_latlng`)
- Added instruction that `sqlQuery` must be a pure SELECT (no INSTALL/LOAD)
- Added table-switching guidance — AI now always calls `resolveTable` when `tableName` is explicitly provided
- Handles quoted/qualified table identifiers safely when normalizing dataset sources
### DeckJsonMap rendering improvements
- Switched to `interleaved: true` by default for GeoArrow layers
- Added `onRenderingError` callback prop and graceful error overlay
- Added heatmap WebGL framebuffer leak workaround (unmount/remount overlay on type change)
- Soft-warning (instead of hard error) when a `@@=column` accessor references a missing column
### Dataset handling
- Panel issue reporting now aggregates errors across all datasets (no longer clears prematurely)
- Added data sampling support with a sampling indicator in the UI
- FROM-clause rewriting uses lookahead instead of `\b` to support quoted table names
### Python server
- Wrapped `get_arrow` in an explicit transaction to prevent "ActiveTransaction" errors during geometry CRS lookup
- Added `_get_arrow_cast_geometry` fallback that converts geometry columns to WKB before Arrow export
- Hardened SQL subquery construction (strip trailing semicolons, escape column names)
### DuckDB connectors
- GeoJSON file loading now falls back to `load('auto')` only when the spatial extension is unavailable, re-throwing genuine parse failures
### Other
- Added `|| !colorScale.domain` guard and `Number.isFinite` checks for empty-domain edge cases in color scales
- Suppressed unhandled promise rejection warnings in `DuckDbSlice` query handles

## Summary by CodeRabbit

* **New Features**
  * Added GeoJSON file upload support across the application
  * Introduced TripsLayer animation with time-based controls
  * Added collapsible color scale legends
  * Implemented fit-to-bounds functionality for deck maps
  * Enhanced layer support (Arc, Heatmap, H3 Hexagon, Trips, Column layers)
  * Added data sampling indicators for large datasets

* **Bug Fixes**
  * Fixed geometry export for Arrow conversion
  * Resolved layer rendering issues
  * Improved handling of missing columns and invalid data

* **Documentation**
  * Updated deck map configuration documentation for interleaved rendering


packages/deck/src/json/layers/SqlroomsTripsLayer.ts
```diff
@@ -0,0 +1,198 @@
...
+ * GeoArrowTripsLayer from @geoarrow/deck.gl-layers@0.3.2 is incompatible with
+ * deck.gl 9.3.x: it passes the raw timestamps typed array via binary attributes,
+ * but deck.gl 9.3's PathLayer/TripsLayer attribute manager fails to initialize
+ * them correctly, producing "Float64Array Error: Float64Array".
+ *
+ * This wrapper extracts the binary data from Arrow vectors and passes it to the
+ * native TripsLayer in a compatible format, properly resolving accessor props
+ * (Arrow Vectors → binary attributes, functions → wrapped with batch context).
+ */
+export class SqlroomsTripsLayer extends CompositeLayer<{
```
ilyabo: I'd avoid using `Sqlrooms` prefix, maybe `DeckTripsLayer`?

But maybe the best would be to submit an issue/PR to @geoarrow/deck.gl-layers

igorDykhta: We should probably upgrade as well to the latest. 
`@geoarrow/deck.gl-layers` has been renamed to `@geoarrow/deck.gl-geoarrow`.
I tried to upgrade quickly, but it only partially worked and requires more work.

ilyabo: ah yes, feel free to upgrade as part of this PR if it works out

ilyabo:
I noticed that the map doesn't seem to respect the selected dashboard's table:

image

ilyabo:
I think we should probably use "Point map" instead of "Scatterplot" in the user-facing UI

image

ilyabo:
Max rows doesn't seem to have any effect, also, we probably shouldn't expose it to the user. I think it should probably be constants depending on the layer type. 

A nice way to deal with it could be:  instead of completely blocking the rendering, we could [sample the data](https://duckdb.org/docs/current/sql/query_syntax/sample) and just show a warning "data is sampled" (and maybe allow users to adjust the max sample size).
 
image

packages/deck/src/prepare/detectGeometryColumn.ts
```diff
@@ -17,14 +17,28 @@ type DetectGeometryColumnOptions = {
...
   return table.schema.fields.map((field) => field.name);
 }
 
-function findCoordinateColumns(table: arrow.Table) {
+/**
+ * Detects longitude/latitude columns in an Arrow table by matching
+ * field names against known patterns (case-insensitive).
+ * Returns the original field names if found, or null.
+ */
+export function findCoordinateColumns(
```
ilyabo, line 38: do you think we should still keep this brittle column sniffing logic at all?  maybe for manual config we should always require the user to select the columns, while the AI assistant can figure it by itself?

igorDykhta: I would say let's keep it for now, and maybe remove it in the next PR.

ilyabo: Should we try to land this one then?

igorDykhta: Actually, I removed this sniffing logic. Looks like with better instructions, the AI is quite good at suplying proper input data.

ilyabo (approved):

python/sqlrooms-server/sqlrooms/server/query.py
```diff
@@ -57,16 +57,65 @@ def retrieve(cache, query, get):
 
 
 def get_arrow(con, sql):
```
ilyabo, line 59: Asked codex about it:

---

Short version: WKB is a reasonable map transport format, but I would be cautious about doing this silently inside generic `get_arrow()`.

In [[PR #700](https://github.com/sqlrooms/sqlrooms/pull/700)](https://github.com/sqlrooms/sqlrooms/pull/700), the fallback addresses a real class of failure: DuckDB/Python Arrow export can need spatial metadata while materializing the Arrow table, and the PR tries to keep the transaction alive, then falls back to `ST_AsWKB(...)` if that still fails. DuckDB documents `ST_AsWKB` as returning a WKB blob, and the Python API’s `to_arrow_table()` is exactly the path SQLRooms uses to ship Arrow to the browser.

**Pros**
- Unblocks map queries that currently fail during Arrow export.
- WKB is a portable, browser-decodable format and SQLRooms’ deck code already understands WKB/`geoarrow.wkb`.
- Avoids relying on DuckDB’s spatial Arrow extension metadata path when that path is the thing failing.
- Matches the convention already used in several map queries: `ST_AsWKB(geom) AS geom`.

**Cons / implications**
- It silently changes the result contract: `GEOMETRY` becomes plain `BLOB`/binary. In my local check with DuckDB `1.5.3`, plain `GEOMETRY` exported as Arrow binary with `ARROW:extension:name = geoarrow.wkb`; `ST_AsWKB` exported as plain binary without that GeoArrow metadata.
- It loses semantic metadata, especially CRS/geometry extension metadata. That may be exactly the metadata DuckDB was trying to look up.
- It makes server/WebSocket results differ from WASM or Node results for the same SQL.
- The fallback re-runs the query. That can be expensive, non-deterministic, or surprising for volatile/table-function queries.
- Wrapping arbitrary SQL as `SELECT * FROM (${sql})` is brittle: trailing semicolons, quoted identifiers, and non-SELECT statements need care. CodeRabbit’s concern here is valid.
- If cached under the original SQL key, the same SQL could cache either native GeoArrow-ish output or fallback WKB depending on whether the first run hit the fallback.

My preference: keep the explicit transaction fix, but don’t make server-wide `get_arrow()` silently rewrite geometries unless we treat it as a narrow last-resort compatibility fallback.

A better direction would be to normalize geometry at the map/data-prep boundary: when a deck dataset needs WKB, generate `ST_AsWKB(geom) AS geom` explicitly and set `geometryEncodingHint: 'wkb'`. That keeps the generic SQL query contract honest while making map behavior reliable. If the server fallback stays, I’d harden it with semicolon stripping, identifier escaping, tests, and ideally restore `geoarrow.wkb` Arrow field metadata after casting.

packages/mosaic/src/charts/chart-types/base-types.ts
```diff
@@ -200,7 +200,11 @@ export type CreateSpecOptions<TSettings = ChartSettings> = {
 };
 
 export function getChartTableReference(dataTable: DataTable): string {
-  return dataTable.table.toString();
+  const {schema, table} = dataTable.table;
+  if (schema && schema !== 'main') {
```
ilyabo: I would revert this change, better use what' s in the main branch here (if you resolve the conflict) :
```
  return getMosaicTableReferenceString(dataTable.table);
```

PR #708, opened 2026-06-10
feat: Add lline chart color selector

dmitriy-kostianetskiy:
image

## Summary by CodeRabbit

* **New Features**
  * Histogram bars can be assigned a custom color via a color selector; histogram settings now include an optional color.
  * Line charts support per-series color selection for Y fields and use those colors in rendering and legend labels.

* **Refactor**
  * Line chart field configuration reorganized into dedicated X-axis and Y-axis selector components.
  * Multi-field selector now accepts a render-prop for per-field customization.

* **UI Components**
  * Added a reusable color selector and a shared default chart color palette; selectors accept optional styling hooks.

* **Style**
  * Minor combobox icon spacing adjustment.

* **Tests**
  * Added tests for color-assignment utility used by line charts.


packages/mosaic/src/constants/chart-colors.ts
```diff
@@ -0,0 +1,13 @@
+// Chart color palette (default palette colors)
+export const CHART_COLORS = [
```
ilyabo: I'd suggest to use `categoricalSchemeColors.Observable10` from `@sqlrooms/color-scales` here

ilyabo (approved):

PR #746, opened 2026-06-21
feat: dashboard maps improvements

igorDykhta:
## Summary

Improvements to dashboard map settings, legends, heatmap handling, and AI instructions.

### Legend improvements

* **Scrollable legends**: CategoricalLegend and SteppedLegend now scroll when exceeding 15 items, with a max of 50 displayed items and a "… and N more" indicator for larger sets
* **Heatmap density legend**: Heatmap layers now show a continuous gradient legend (Low → High) using the active color scheme
* **No stale legends on type switch**: Switching from a point layer (with color scale) to a heatmap no longer keeps the point layer's legend

### Heatmap color scheme

* Default heatmap `colorRange` is now YlOrRd (matching deck.gl's built-in default)
* When `colorRange` is absent, the rendering pipeline injects the default so the UI scheme selector and actual colors are always in sync
* `updateTriggers` are set to force re-aggregation when `colorRange` changes

### Map settings panel

* **Unit-aware sliders**: Line width and point radius sliders now detect the layer's unit setting (`pixels` vs `meters`), adjusting labels (px/m) and ranges (1–50 for pixels, 1–10000 for meters) accordingly — no more silent unit overrides
* **Color scale auto-detection**: Switching the color field now auto-detects `sequential` vs `categorical` scale type based on column type
* **Generic column selector for color field**: Replaced `ColumnSelector.Quantitative`/`ColumnSelector.Categorical` with a generic `ColumnSelector`, since type detection is now automatic
* **No black points on color scale clear**: `clearDeckMapLayerColorScale` for `getFillColor` now resets to default blue instead of deleting the accessor

### Geometry column and fitToData

* `_sqlroomsBinding.geometryColumn` is now explicitly included in generated layer configs
* `fitToData` is now set for geometry-column-backed layers (not just lon/lat)
* `setDeckMapLayerGeometryColumn` keeps `fitToData.geometryColumn` in sync when the user changes the geometry column in settings

### Layer availability

* `filterUnavailableLayers` now passes through layers when dataset status is not `error` (previously required `ready`), reducing layer flickering during cross-filtering

### AI instructions

* Added CRITICAL `geometryColumn` rule (must match the exact WKB output alias)
* Added HEATMAP instruction (omit `colorRange`, let the default apply)
* Added BASEMAPS instruction (avoid Mapbox styles, use CARTO)
* Strengthened `fitToData` format instruction (must be flat object, not nested by dataset ID)
* Added defensive normalization in `normalizeAiMapConfig` to auto-fix malformed nested `fitToData` structures from AI

### Removed (dead import cleanup)

* Removed unused `TableColumn` type import from MapSettings


packages/deck/src/MapSettings.tsx
```diff
@@ -68,6 +68,46 @@ function extractTableFromSqlQuery(
...
 }
 
+/**
+ * Extracts column aliases from a SQL SELECT clause (e.g. `... AS geom`).
+ * Returns names of columns that are produced by the query but may not
+ * exist in the raw source table.
+ */
+function extractSelectAliases(sqlQuery: string | undefined): string[] {
+  if (!sqlQuery) return [];
+  const selectMatch = sqlQuery.match(/\bSELECT\b(.+?)\bFROM\b/is);
```
ilyabo: oh my, is there no other way except for parsing SQL? this is very brittle.
e.g. if the query comes from the model maybe we could instruct it to provide the name as additional output

ilyabo: AI suggested this, I haven't tested but maybe you can take a look:
https://github.com/sqlrooms/sqlrooms/pull/768


> Yes, we can avoid it. I removed extractSelectAliases() from [MapSettings.tsx](/Users/ilya/Workspace/sqlrooms/packages/deck/src/MapSettings.tsx) and instead pass the loaded map datasetStates into MapSettingsPanel from [dashboard.tsx](/Users/ilya/Workspace/sqlrooms/packages/deck/src/dashboard.tsx). The settings panel now uses the actual resolved Arrow table schema for the active dataset, so computed columns like AS geom come from DuckDB’s query result rather than a regex over SQL text.

igorDykhta: I removed the generated column logic. It makes sense to implement it in a separate isolated PR. I tried several approaches, but each time I ran into issues and couldn't get everything working correctly.

ilyabo (approved):

PR #799, opened 2026-07-07
feat: worksheet dashboard

dmitriy-kostianetskiy:
## Summary by CodeRabbit

* **New Features**
  * CLI AI and command workflows now operate on block documents, with updated support for dashboards, data tables, maps, and HTML apps.
  * Block-related actions now return clearer IDs and status details for created or updated content.

* **Bug Fixes**
  * Improved map handling so existing maps and panels can be updated more reliably.
  * Import flow now uses the current artifact consistently when adding table content.

* **Chores**
  * Updated the AI package version across multiple workspace packages.

### DIFF
--- a/packages/ai-core/package.json
+++ b/packages/ai-core/package.json
@@ -47,7 +47,7 @@
     "@sqlrooms/room-store": "workspace:*",
     "@sqlrooms/ui": "workspace:*",
     "@sqlrooms/utils": "workspace:*",
-    "ai": "6.0.177",
+    "ai": "^6.0.177",
     "immer": "^11.0.1",
     "lucide-react": "^0.556.0",
     "react-markdown": "^10.1.0",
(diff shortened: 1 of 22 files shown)


packages/ai-core/package.json
```diff
@@ -47,7 +47,7 @@
     "@sqlrooms/room-store": "workspace:*",
     "@sqlrooms/ui": "workspace:*",
     "@sqlrooms/utils": "workspace:*",
-    "ai": "^6.0.154",
+    "ai": "6.0.177",
```
ilyabo: I think we shouldn't use the precise version in the published packages, it might be to much of an constraint to the consumers, likely unnecessary. I would keep using "^", so "^6.0.177" should be fine.

ilyabo (approved):

PR #816, opened 2026-07-16
fix: rollback chat context badge selector

bdjulbic:
Screenshot 2026-07-16 at 09 54 39

Rollback chat badge selection so it shows which Map is active

## Summary by CodeRabbit

* **New Features**
  * Added index-aware “main” context chip behavior: the first selected item is treated as the main chip and is visually highlighted.
  * Main chips display the icon and truncated title with a “Primary context” tooltip, while non-main chips remain draggable and show the same icon/title alongside removal.
* **Bug Fixes / UI Updates**
  * Updated context badge logic so the “Selected” badge is shown based only on selection state.
  * The “Running” badge is no longer rendered by the context item component.

### DIFF
--- a/packages/ai-core/src/components/context/ContextSelectorSortableChip.tsx
+++ b/packages/ai-core/src/components/context/ContextSelectorSortableChip.tsx
@@ -45,7 +45,8 @@ export const ContextSelectorSortableChip: FC<
         {main ? (
           <Tooltip>
             <TooltipTrigger asChild>
-              <span
+              <button
+                type="button"
                 className="flex min-w-0 items-center gap-1"
                 aria-label={`${item.title} is the primary context item`}
               >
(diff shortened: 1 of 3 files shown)


packages/ai-core/src/components/context/ContextSelectorSortableChip.tsx
```diff
@@ -44,12 +46,17 @@ export const ContextSelectorSortableChip: FC<
         >
           <GripVerticalIcon className="h-2.5 w-2.5" />
         </button>
-        <span className="flex min-w-0 items-center gap-1">
+        <button
```
ilyabo: I would not make it clickable again to reorder, I think it was confusing

packages/ai-core/src/components/context/ContextSelectorSortableChip.tsx
```diff
@@ -32,7 +34,7 @@ export const ContextSelectorSortableChip: FC<
       }}
     >
       <UiBadge
```
ilyabo, line 41: We should also add a tooltip saying "Primary context item"

packages/ai-core/src/components/context/ContextSelectorSortableChip.tsx
bdjulbic: removed the click functionality. Now user can only drag to the front

ilyabo (approved):

PR #817, opened 2026-07-16
feat(mosaic): per-column summary kind overrides for DataTableExplorer

NickCrews:
## Problem

`DataTableExplorer`'s `columns` option is all-or-nothing: a column is either loaded — with its histogram/category summary and cross-filter UI — or excluded from the table entirely. There is no way to:

- show a column's values while disabling its summary and filtering (e.g. `Latitude`/`Longitude`, where a histogram and brush add noise and query load but no insight), or
- override the Arrow-type-driven summary kind (e.g. a numeric code column that reads better as category buckets than a histogram).

The summary kind decision (`histogram` / `category` / `unsupported`) was hard-coded from the Arrow type in three separate places (`useDataTableExplorerColumns`, `connectDataTableExplorerSummaryClients`, `createEmptySummaryState`), so there was no seam to hook into. Hiding the summary UI from outside wouldn't help either: the summary Mosaic clients would still be created and still run their queries.

## Solution

Add one deep option instead of a per-column config object:

```tsx
<DataTableExplorer
  tableName={dataTable.table}
  selection={brush}
  getColumnKind={(field) => {
    if (field.name === 'Latitude' || field.name === 'Longitude') return 'none';
    return 'auto';
  }}
>
```

- `resolveDataTableExplorerColumnKind(field, getColumnKind?)` is now the single decision point for a column's kind; the three call sites consult it instead of re-deriving from the Arrow type.
- `'none'` renders the column with no summary cell and **creates no summary clients**, so no summary queries run and the column doesn't participate in cross-filtering. Sorting and row rendering are unchanged.
- `'category'` / `'histogram'` force a specific summary. Incompatible requests (histogram on a string column, any summary on binary/geometry) fall back to the type-driven default rather than producing broken queries.
- Resolved kinds are **value-memoized** in `useDataTableExplorer`, so an inline resolver lambda can't tear down and reconnect every summary client on each render — the client lifecycle only restarts when a column's resolved kind actually changes.
- For custom headers built on `useDataTableExplorer`, the default summary cells are now exported: `DataTableExplorerSummaryCell`, `DataTableExplorerHistogramSummaryCell`, `DataTableExplorerCategorySummaryCell`, plus the resolver itself.

Design notes: `columns` stays a pure projection; render concerns (widths, labels, custom cells) stay in the compound-component layer. The resolver is a function, so persisted surfaces (dashboard panels / block documents) that want this can layer a serializable `Record<string, kind>` on top later — deliberately not included here. A read-only summary mode ("show histogram, disable brushing") is a distinct axis and also deliberately out of scope.

## How to test

1. `pnpm install && pnpm build`
2. Unit tests for the resolver (default kinds, `'auto'`, `'none'` always honored, incompatible overrides rejected):
   ```sh
   cd packages/mosaic && pnpm test
   ```
   24 suites / 136 tests pass, including the new `resolveDataTableExplorerColumnKind` block in `__tests__/data-table-explorer.utils.test.ts`.
3. See it live in the deckgl-mosaic example, which now sets `Latitude`/`Longitude` to `'none'`:
   ```sh
   pnpm dev deckgl-mosaic-example
   ```
   In the **Earthquake Profiler** table at the bottom:
   - `Latitude` / `Longitude` show name + type badge only — no histogram, no brush. Clicking the header still sorts.
   - Every other column is unchanged (`DateTime`/`Magnitude` histograms, `MagType`/`Source` category buckets).
   - Click the `Md` bucket under `MagType`: rows filter (≈45,330 of 54,936), all summaries update, and the `'none'` columns stay summary-less.
   - Optional: temporarily return `'category'` for `Depth` in `EarthquakeProfiler.tsx`'s `getColumnKind` to see a numeric column forced to category buckets, with bucket-click filtering working.

All three steps verified locally (macOS, headless Chromium screenshots for step 3).

### DIFF
--- a/examples/deckgl-mosaic/src/components/filters/EarthquakeProfiler.tsx
+++ b/examples/deckgl-mosaic/src/components/filters/EarthquakeProfiler.tsx
@@ -1,8 +1,28 @@
 import {useDataTable} from '@sqlrooms/duckdb';
-import {DataTableExplorer} from '@sqlrooms/mosaic';
+import {
+  DataTableExplorer,
+  type DataTableExplorerColumnKindOverride,
+} from '@sqlrooms/mosaic';
 import {cn} from '@sqlrooms/ui';
+import type {Field} from 'apache-arrow';
 import {FC, useMemo} from 'react';
 import {useRoomStore} from '../../store';
+
+// Demonstrates per-column summary-kind overrides: coordinate columns are
+// shown without summaries (they stay sortable but don't cross-filter), and
+// MagType keeps its type-driven category buckets. Columns not listed here
+// keep the default Arrow-type-driven behavior via 'auto'.
+const COLUMN_KIND_OVERRIDES: Record<
+  string,
+  DataTableExplorerColumnKindOverride
+> = {
+  Latitude: 'none',
+  Longitude: 'none',
+};
+
+function getColumnKind(field: Field): DataTableExplorerColumnKindOverride {
+  return COLUMN_KIND_OVERRIDES[field.name] ?? 'auto';
+}
 
 type EarthquakeProfilerProps = {className?: string};
 
@@ -35,6 +55,7 @@ export const EarthquakeProfiler: FC<EarthquakeProfilerProps> = ({
         pageSize={25}
         selection={brush}
         tableName={dataTable.table}
+        getColumnKind={getColumnKind}
       >
         <div className="flex items-center justify-between gap-4 px-3 py-2">
           <div>
--- a/examples/deckgl-mosaic/src/components/filters/EarthquakeProfiler.tsx
+++ b/examples/deckgl-mosaic/src/components/filters/EarthquakeProfiler.tsx
@@ -2,9 +2,9 @@ import {useDataTable} from '@sqlrooms/duckdb';
 import {
   DataTableExplorer,
   type DataTableExplorerColumnKindOverride,
+  type DataTableExplorerOptions,
 } from '@sqlrooms/mosaic';
 import {cn} from '@sqlrooms/ui';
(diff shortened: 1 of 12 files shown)


ilyabo:
Thanks for working on this! I agree that the current all-or-nothing behavior is limiting, and adding a way to control the summary strategy per column seems like a useful extension point.

I’m less convinced by the motivating latitude/longitude example as currently framed. `DataTableExplorer` is a generic table preview/profiling component and generally shouldn’t know the semantic meaning of columns in advance. Latitude/longitude histograms and brushing can still reveal range, clustering, precision issues, and outliers. An application may choose to suppress them—for query cost, layout, or because a map provides a better interaction—but that is an application policy rather than a general property of coordinate columns.

For that reason, I would avoid changing the earthquake demo. It makes a subjective application choice look like the recommended behavior for coordinate columns. The resolver can be demonstrated through focused tests and documentation instead.

The resolver itself makes sense to me. I would consider naming it `getColumnSummaryKind` or `resolveColumnSummaryKind`, since it controls the summary/query strategy rather than the semantic “kind” of the column.

A couple of implementation concerns:

- When the resolved kinds change, all summary clients are destroyed, but existing selection clauses created by those clients do not appear to be removed. Changing a selected column to `none` could therefore leave a hidden filter active with no corresponding summary UI.
- The memoization key concatenates column names and kinds using `:` and `|`, both of which may occur in column names. I would use `JSON.stringify(columnKindEntries)` rather than SQL identifier escaping here; this is a serialization key, not SQL syntax.

Overall, I support adding a summary-kind resolver. I’d suggest removing the demo-specific coordinate policy, tightening the naming, and addressing the lifecycle and memoization issues above.

PR #830, opened 2026-08-03
refactor(ai-core): deepen chat rendering API

ilyabo:
## Summary

- replace the wide custom `Turn` prop bag with pre-wired semantic regions
- resolve default rendering deterministically without global module registration
- route top-level and nested hoisted output through the configured `HoistedOutput` slot
- normalize `shouldHoist` state and support memo/forwardRef-style renderers
- narrow the new public export surface and update the customization guide

## Why

This is a stacked follow-up to #828. The presentation-slot direction is sound, but the first API shape exposed internal search, hoisting, and render-plan details to every full-turn layout. It also allowed the default recipe to bypass advertised slots and depended on import-order initialization.

Custom turn layouts now receive small turn metadata plus zero-prop regions (`Prompt`, `Timeline`, `Activity`, `Response`, `HoistedOutputs`, `Summary`, and `Actions`). SQLRooms remains responsible for binding semantic data, search ids, actions, and hoisting decisions.

## Developer impact

A full custom layout becomes a short composition of regions rather than a reimplementation of turn rendering. Partial leaf-slot overrides continue to merge and now apply consistently to hoisted output. Since #828 is not merged, this refines the new API before release rather than introducing a migration for an existing public contract.

## Validation

- `pnpm build`
- `pnpm knip`
- `pnpm typecheck`
- `pnpm check-circular-deps`
- `pnpm test`
- `pnpm --filter @sqlrooms/ai-core lint`
- `pnpm check:typedoc-links`


lixun910 (approved):
LGTM! 👍

packages/ai-core/README.md
```diff
@@ -229,79 +228,29 @@ You can combine any of these in one map, e.g.
...
-          .join('\n')}
-        canFork={false}
-      />
+    <article data-turn-id={turn.id} aria-busy={!turn.isCompleted}>
+      <Prompt />
+      <Activity />
+      <Response />
+      <HoistedOutputs />
+      <Summary />
+      <Actions />
```
lixun910, line 253: @chatgpt-codex-connector I think the content should be an element, not a component. E.g.: If the library handed you a fresh anonymous component on each presentation rebuild, React would treat it as a new type every render → remount charts/tools → lost local state, and in this app that can loop with query-driven store updates and causing flickering.

PR #856, opened 2026-08-19
feat(ai-core): expose active status rendering

### LINKED ISSUE #854: feat(ai-core): export chat active-status derivation and allow hosts to own the progress indicator
### Context

`Chat.Rendering` (added in #828) established a clear division: SQLRooms owns the semantic model for a chat turn, and host applications may own presentation. That division is honoured thoroughly for turn content — every region and leaf slot is overridable, and the semantic data behind each one is exposed.

There is one place where it isn't: the **run's current active status** — the "what is the agent doing right now" indicator shown while a run is in flight.

### Problem

`getChatActiveStatus(messages, behavior)` derives exactly the right semantic value for this — `{key, label, kind: 'tool' | 'approval' | 'model'}` — and it already honours `ToolRenderBehavior`, so a host's `getActivityLabel` / `getToolDisplayName` flow into it correctly. `ChatActiveStatus` renders it, and `ChatMessagesContainer` renders that internally whenever the run is active.

But neither the function, the component, nor the `ChatActiveStatusInfo` type is exported from the `@sqlrooms/ai-core` barrel, and the package's `exports` map is limited to `"."` and `"./devtools"`, so a deep import isn't available either. The net effect is that the *semantics* of active status are public in spirit but private in practice, while `AiThinkingDots` — a pure presentation primitive with no semantics at all — **is** exported. That's inverted relative to the rest of the API.

A host that wants a different progress indicator (a different animation, different placement, an elapsed-time treatment of its own, or different dismissal timing relative to the first answer token) currently has to reimplement the derivation. That means duplicating message-part walking, the tool state machine across `output-available` / `output-error` / `output-denied` / `approval-requested` / `approval-responded`, the visible-progress predicate, and the elapsed-time bookkeeping — logic that will keep evolving upstream, so any copy drifts silently and ends up disagreeing with the activity timeline it sits beneath.

### Proposal

**1. Export the existing symbols from the barrel.** `getChatActiveStatus`, `ChatActiveStatus`, and the `ChatActiveStatusInfo` type. Purely additive, no behaviour change, and the code is already written and documented. This lets a host compose the derivation with its own presentation — including with `AiThinkingDots`, which is already public.

**2. Let a host suppress or replace the built-in indicator.** Because `ChatMessagesContainer` renders `ChatActiveStatus` internally, exporting the derivation alone means a host that renders its own indicator shows two. Some way to opt out is needed for the export to be usable. Options, in rough order of how well they fit the existing API:

- An `ActiveStatus` slot on the `Chat.Rendering` recipe, defaulting to the current component. This matches how every other region is already customised, keeps placement upstream, and needs no new prop — probably the most consistent option.
- A prop on `Chat.Messages` to disable the built-in render, leaving the host to place its own.
- A boolean/`ReactNode` override prop for the indicator specifically.

The slot approach seems most in keeping with #828, since it treats active status as one more presentation region over a semantic model SQLRooms owns — but any of the three unblocks the use case.

### Why this is generally useful

Any application embedding the chat with its own visual language will want its progress indicator to match that language, and long-running agents make this indicator one of the most-looked-at elements in the UI. It's also the one piece of run state that stays visible when a host chooses to hide the reasoning trace entirely, which makes it the sole progress signal in that configuration. Right now that's the one thing a host cannot restyle.

Happy to open the PR for whichever shape you prefer.

ilyabo:
## Summary

- export `getChatActiveStatus`, `ChatActiveStatus`, `ChatActiveStatusInfo`, and the slot props from `@sqlrooms/ai-core`
- add an `ActiveStatus` slot to `Chat.Rendering`, with the existing SQLRooms indicator as the default
- pass presentation-ready status semantics to the slot in both session and local-agent chat runtimes
- document custom and suppressed active-status presentations

## Why

Hosts can already customize every other chat presentation region, but the in-flight progress indicator was rendered internally. This forced applications with their own visual language to duplicate active-status derivation or show two indicators.

The new slot keeps status derivation in SQLRooms while allowing hosts to replace, suppress, or relocate its presentation.

Closes #854

## Validation

- `pnpm build`
- `pnpm --filter @sqlrooms/ai-core build`
- `pnpm --filter @sqlrooms/ai-core typecheck`
- `pnpm --filter @sqlrooms/ai-core lint` (passes with two pre-existing warnings)
- `volta run --node 24 pnpm --filter @sqlrooms/ai-core test` (31 suites, 178 tests)
- `pnpm check-circular-deps`
- full pre-push typecheck and test hooks under Node 24

## Summary by CodeRabbit

- **New Features**
  - Added customizable active-status indicators for current chat runs, including elapsed-time presentation.
  - Added support for replacing or suppressing the active-status display through chat rendering customization.
  - Exposed helpers and types for rendering active status consistently in custom interfaces.

- **Documentation**
  - Expanded chat rendering documentation with active-status customization examples and guidance.

### DIFF
--- a/packages/ai-core/README.md
+++ b/packages/ai-core/README.md
@@ -117,6 +117,7 @@ order, what is hoistable / running / completed), then asks the nearest
 
 | Slot            | Role                                                          |
 | --------------- | ------------------------------------------------------------- |
+| `ActiveStatus`  | Current in-flight run status and elapsed-time presentation    |
 | `Turn`          | Full turn layout recipe — composes pre-wired semantic regions |
 | `Prompt`        | User prompt chrome for the turn                               |
 | `Activity`      | Collapsible / status chrome around in-progress work           |
@@ -228,6 +229,27 @@ function AppReasoning({text, isRunning}: ChatReasoningProps) {
 
 You can combine any of these in one `components` map by providing both the
 `Prompt` and `Reasoning` entries.
+
+**Active run status** — change or suppress the progress indicator without
+reimplementing chat-run state derivation. The exported `getChatActiveStatus`
+helper is also available when the indicator needs to be placed outside
+`Chat.Messages`:
+
+```tsx
+import {Chat, type ChatActiveStatusProps} from '@sqlrooms/ai-core';
+
+function AppActiveStatus({status}: ChatActiveStatusProps) {
+  return <MyProgressIndicator label={status.label} />;
+}
+
+<Chat.Rendering components={{ActiveStatus: AppActiveStatus}}>
+  <Chat.Messages />
+</Chat.Rendering>;
+```
+
+Set `ActiveStatus` to a component that returns `null` when the host owns the
+indicator's placement entirely. For that case, call `getChatActiveStatus` with
+the current messages and `ToolRenderBehavior` to reuse the same status model.
 
 `TextOutput` receives `isAnswer=true` only for text that is the final message
 part. Planning text followed by tool activity remains regular response text.
--- a/packages/ai-core/__tests__/ChatRenderingContext.test.tsx
+++ b/packages/ai-core/__tests__/ChatRenderingContext.test.tsx
@@ -4,13 +4,24 @@
 import {act} from 'react';
 import {createRoot} from 'react-dom/client';
 import {TransformStream} from 'node:stream/web';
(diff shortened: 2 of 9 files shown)


mihajlo-stojkovic (approved):
Approving, this is exactly the shape I was hoping for, thanks for turning it around so quickly 🚀

PR #858, opened 2026-08-19
chore(kepler): upgrade to 3.3.0-alpha.7

ilyabo:
## Summary

- upgrade all direct `@kepler.gl/*` dependencies from `3.3.0-alpha.0` to `3.3.0-alpha.7`
- align React Intl and styled-components with the upgraded Kepler packages
- adapt custom legend and drag-and-drop integrations to the alpha.7 exports and grouped layer-order types
- add test-only ESM interop shims for Kepler's CommonJS `global/window` and styled-components dependencies

## Impact

SQLRooms' Kepler package and example now build against `3.3.0-alpha.7` while retaining custom legend and layer drag-and-drop behavior.

## Validation

- `pnpm build` — 49/49 package builds passed
- `pnpm --filter @sqlrooms/kepler test --runInBand` — 2 suites, 9 tests passed
- `pnpm --filter @sqlrooms/kepler lint` — passed with 14 pre-existing warnings and no errors
- `pnpm --filter kepler-example build` — passed
- pre-push `turbo typecheck` — 54/54 tasks passed
- pre-push circular dependency check — passed

## Known unrelated issue

The repository-wide pre-push test stage fails in `@sqlrooms/room-shell` on Node 22 because Jest attempts to synchronously load the ESM-only `vscode-languageserver-types` entry and reports that Node 24.9+ is required. The relevant lockfile versions are unchanged from `main`, and the focused Kepler tests pass.

## Summary by CodeRabbit

* **New Features**
  * Improved drag-and-drop layer ordering across groups, including moving layers between groups and to the root.
  * Updated map legends to reflect custom layer ordering, visibility states, and split-map controls.
  * Added support for React 19 and updated Kepler integrations.

* **Bug Fixes**
  * Improved handling of edge cases when reordering layers, including canceled drops, empty groups, and end-of-list placement.

* **Tests**
  * Added comprehensive coverage for layer reordering, drag-and-drop routing, and legend ordering.


igorDykhta:
@ilyabo Please don't rush with this ;)

PR #866, opened 2026-08-20
fix(artifacts,ai-config): give Node consumers a React-free entry point

ilyabo:
## Summary

Two small changes so a Node consumer can import the artifact data model without parsing the ESM build or reaching into `src/…` internals.

## `ai-config`: fix an inferred `never[]`

`const newParts = []` in the v0.25.0 migration infers `never[]`, so both `newParts.push(...)` calls below it fail:

```
src/migration/AnalysisSession-v0.25.0.ts(104,21): error TS2345:
  Argument of type '{ type: string; text: unknown; }' is not assignable to parameter of type 'never'.
src/migration/AnalysisSession-v0.25.0.ts(129,23): error TS2345: ...
```

The package itself compiles because the file is only typechecked in isolation. But any project that pulls this source into its own program — a monorepo host mapping the package to source, say — inherits two errors it cannot fix from the outside. One annotation resolves both.

## `artifacts`: add a React-free `./config` subpath

`ArtifactMetadata`, `ArtifactsSliceConfig`, `ArtifactType` and `ArtifactSessionLink(Schema)` only import zod. But the package root also re-exports slices and components, so it pulls React, and today the only way to reach those shapes from Node is to import an internal module path.

That's brittle in a way that has already bitten: `ArtifactSessionLink` moved out of `ArtifactsSliceConfig`, and importers that had reached in broke silently — the symbol resolved to `undefined` rather than failing at import.

`src/ai/` is already React-free and needs no change. This gives the config layer the same property, so both are usable from a test runner, a migration script, or a server-side tool.

## Why no CJS build

A dual build would fix the "cannot parse ESM" half, but not the reason a Node consumer can't use the root entry in the first place — React. Subpaths address the actual cause, so I've left the build alone.

## Verification

- Both packages typecheck; `ai-config` tests pass.
- `tsc` emits `dist/config.js` and `dist/config.d.ts`, so the subpath resolves from the published build.
- Downstream check: a host that was carrying **three** hand-written re-export shims to work around exactly this deleted all three and kept its suite green — 38 suites / 441 tests, typecheck clean.

Note: `artifacts`' own jest suites fail to parse on `main` and still do here. Pre-existing and deliberately untouched.

## Summary by CodeRabbit

* **New Features**
  * Added a public configuration export for artifact metadata, slice settings, artifact types, and session links.
  * Configuration utilities can now be accessed without loading React-specific functionality.

* **Refactor**
  * Improved type clarity in legacy analysis-session migration logic without changing runtime behavior.

### DIFF
--- a/packages/ai-config/src/migration/AnalysisSession-v0.25.0.ts
+++ b/packages/ai-config/src/migration/AnalysisSession-v0.25.0.ts
@@ -96,7 +96,7 @@ function migrateStreamMessage(streamMessage: unknown) {
   }
 
   const parts = (streamMessage as {parts: Record<string, unknown>[]}).parts;
-  const newParts = [];
+  const newParts: Record<string, unknown>[] = [];
 
   for (const part of parts) {
     if (part.type === 'text') {
--- a/packages/artifacts/README.md
+++ b/packages/artifacts/README.md
@@ -154,6 +154,37 @@ The artifact shell still owns workspace metadata such as id, title, tabs,
 current selection, and AI context. The stateful block definition owns the
 feature-specific rendering and backing-state lifecycle.
 
+## Entry Points
+
+| Import                       | Contains                                                                                                            | Pulls React |
+| ---------------------------- | ------------------------------------------------------------------------------------------------------------------- | ----------- |
+| `@sqlrooms/artifacts`        | Slices, components, artifact types — the full package                                                               | Yes         |
+| `@sqlrooms/artifacts/config` | Serializable shapes only: `ArtifactMetadata`, `ArtifactsSliceConfig`, `ArtifactType`, `ArtifactSessionLink(Schema)` | No          |
+| `@sqlrooms/artifacts/ai`     | Assistant tools for artifact context                                                                                | No          |
+
+Prefer `@sqlrooms/artifacts/config` when you only need the persisted data model
+and want to stay out of the React dependency — a test runner, a config
+migration, or server-side code:
+
+```ts
+import {
+  ArtifactMetadata,
+  ArtifactSessionLinkSchema,
+} from '@sqlrooms/artifacts/config';
+
+const link = ArtifactSessionLinkSchema.parse(row);
+```
+
+Import these shapes from `/config` rather than from an internal module path such
+as `@sqlrooms/artifacts/dist/ArtifactsSliceConfig`. The subpath is the supported
+surface and will keep working when the underlying modules are reorganized;
+internal paths have moved before.
+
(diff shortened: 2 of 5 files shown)


ilyabo:
Thanks both — addressed three of the four, and pushed `868be3a`.

**Documented the entry point** (Codex). README now has an Entry Points table saying what each entry contains and whether it pulls React, with explicit guidance to import the data model from `/config` rather than an internal path. Fair catch: leaving it undocumented would have left Node consumers on exactly the internal paths this change exists to replace.

**Documented the type exports** (CodeRabbit). `ArtifactType`, `ArtifactMetadata` and `ArtifactsSliceConfig` had no TSDoc anywhere, so I added it at the definitions rather than only on the aliases — that way the root entry benefits too. Verified it survives into the emitted `.d.ts`. The `*Type` aliases now also say why they exist.

**Node ESM resolution** (Codex) — the observation is correct, and it made me fix an overclaim of mine rather than the code. `dist/config.js` does emit `from './ArtifactsSliceConfig'` extensionless, so bare `node` cannot resolve it. But `dist/index.js` emits the identical pattern: the package builds with `moduleResolution: "bundler"`, so this is true of `.` and `./ai` today and is not introduced here. Making the published files resolvable by native `node` means changing the shared TS preset for every entry point in the repo, which is well outside a 31-line change.

What was wrong was my docblock implying otherwise. Both it and the README now state plainly that all entries target bundlers/transpilers, and that being React-free is what makes `/config` and `/ai` usable from Node *toolchains* — not native `node` resolution. Happy to open a separate issue for the package-wide question if that is useful.

**Smoke test for the published subpath** (CodeRabbit) — skipped, and worth explaining because the two findings collide. A runtime test against the built package is not currently writable:

1. `packages/artifacts` jest cannot run *any* test on `main` — all four suites fail to parse, on `room-store/dist/index.js:5 SyntaxError: Unexpected token 'export'`. I confirmed this is pre-existing by stashing my changes. A new test there would be added coverage that never executes, which is worse than none.
2. Importing the built entry from plain `node` is precisely what the extensionless emit prevents — so the requested test would fail for the reason above, not because of anything it is meant to catch.

What does hold the guarantee meanwhile: `tsc` emits `dist/config.js` + `dist/config.d.ts` so declarations resolve, the entry is structurally React-free (it imports only zod and two sibling schema modules), and downstream a host that was carrying three hand-written re-export shims for this deleted all three and kept its suite green — 38 suites / 441 tests, typecheck clean. Glad to add the smoke test as a follow-up once the package's jest setup can run.

PR #869, opened 2026-08-20
fix(deck): reject lon/lat maps that never produce geometry

igorDykhta:
## Summary
- Reject point/heatmap configs whose SQL selects lon/lat columns but never produces geometry. Authored SQL is not rewritten.
- Config issues carry `repair` text for AI retry; the map overlay shows only `message`.
- Config and render errors share a collapsible overlay.

## Test plan
- Column-only lon/lat heatmap is rejected with `ST_Point` repair text
- Table-only and `SELECT *` lon/lat maps still pass
- Invalid config shows the overlay instead of a raw "Invalid map configuration" string


ilyabo:
AI feedback (with hints from me):


---

Thanks for addressing this failure mode. The improved repair text and error overlay look useful, but I don’t think we should use the new SQL inspection as a hard validation gate.

`getOuterSelectList` / `sqlIsLonLatProjectionMissingGeometry` amount to a partial SQL parser. They cannot reliably determine a query’s output schema: CTEs, comments, quoted identifiers, aliases, `SELECT *`, DuckDB extensions, and expression types can all produce false positives or negatives. For example, a valid query whose outer `SELECT` adds geometry after a CTE may be rejected because the code examines the first inner `SELECT`. Conversely, merely mentioning `geom` in an expression can make an invalid query pass.

It is also worth separating two possible problems:

- If invalid coordinates are breaking viewport fitting, the simpler fix is to filter the bounds query:

  ```sql
  WHERE longitude BETWEEN -180 AND 180
    AND latitude BETWEEN -90 AND 90
    AND isfinite(longitude)
    AND isfinite(latitude)
  ```

- If the rendered layer lacks point geometry, filtering `fitToData` will not help. The bounds query is separate from layer preparation and does not provide `getPosition`.

For the missing-geometry case, I suggest:

1. Keep synchronous config validation structural only.
2. Remove the semantic SQL regex parsing.
3. Use the existing DuckDB `DESCRIBE SELECT * FROM (<compiled source>)` step to check the actual output columns and types. If validation must happen before writing the resource, expose this as an async preflight with `executeSql`; otherwise report it from dataset preparation with the same agent repair guidance.
4. Longer term, make lon/lat-derived geometry explicit in the config—either as a first-class position binding or structured derived geometry. The preparation layer could then build a GeoArrow point vector directly from the longitude/latitude columns, without generating WKB and decoding it again.

The same coordinate-validity predicate should ideally be shared by rendering and fitting so invalid rows are handled consistently.

So I agree that the failure should be diagnosed, but I think the SQL regex gate should be replaced before merge. If the motivating bug was actually viewport blow-up from invalid coordinates, then this PR is solving a different and substantially more complicated problem.

---

There may be a broader spatial-contract issue here. This validation only handles missing geometry for the lon/lat case. A geometry column can exist but still be unusable—for example, projected EPSG:3857 coordinates passed to a longitude/latitude map. Filtering fitToData ranges would not fix that and could silently hide the entire dataset.
Could we instead define a unified spatial preparation step for coordinate columns and geometry columns? It would resolve or require the source CRS, transform to the map’s expected coordinate system, validate finite/range/type constraints, and provide the canonical geometry used by both rendering and fitting. Row-level bad coordinates could be dropped with a warning; an unknown CRS or dataset-wide range mismatch should be an actionable error rather than silently filtered.
This would solve the motivating lon/lat problem without parsing SQL and would also cover existing geometry columns consistently.


I don’t think this PR needs to implement the complete CRS-aware spatial preparation design. That is broad enough for a focused follow-up. However, I also would not merge the SQL regex as an interim solution. For this PR, could we keep the improved error/repair plumbing and derive the missing-geometry error from actual dataset preparation? A follow-up can then introduce the unified lon/lat/geometry binding and CRS-normalization pipeline without having to unwind a new SQL-parsing layer.

PR #908, opened 2026-09-08
fix(cli): restore legacy workspaces without overwriting failed loads

ilyabo:
Opening an older CLI workspace can fail because saved AI/artifact associations still use `createdAt` and `linkType`, while the current schema requires `linkedAt`. That failure aborts restoration of every slice, and subsequent autosaves can replace the saved documents and maps with empty defaults.

Migrate legacy associations and pinned artifact IDs before validating the workspace. Keep saving disabled until restoration and post-load synchronization succeed, and show a persistent error if loading fails. After successful restoration, persist migrations and retained startup changes when they differ from the loaded snapshot. Existing document-type migrations continue to apply.

Validation:

- 18 regression tests passed across `migrateCliPersistedWorkspace.test.ts` and `serverApi.test.ts`, including migration idempotence, canonical-value precedence, failed restoration followed by autosave/final flush, successful retry, new-workspace saving, retained startup changes during delayed loading, and post-load synchronization.
- `pnpm --filter sqlrooms-cli-app typecheck` passed; formatting and `git diff --check` passed.
- Repository pre-push checks passed: Knip, workspace typechecking, circular-dependency checks, and the full test suite. The first full run hit Python server-startup timeouts; all 28 server tests passed in isolation, and the subsequent complete push hook passed.
- Validated an affected snapshot against all 19 current state schemas and verified a separate recovered DuckDB copy retained 26 block documents, 72 map configurations, and 42 chat associations. Document bodies and map configurations were unchanged. No user database files are included in this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Legacy workspaces now restore AI/artifact links and pinned artifacts correctly when loaded.
  - Workspace data is protected from being overwritten while loading or if restoration fails.
  - Loading errors now display a warning and keep the existing saved workspace intact.
  - Saving becomes available again after successful workspace restoration and validation.
  - Startup changes and successful migrations are now saved automatically after restoration.

### DIFF
--- a/apps/sqlrooms-cli-ui/README.md
+++ b/apps/sqlrooms-cli-ui/README.md
@@ -222,6 +222,8 @@ to `linkedAt`, removes the obsolete `linkType`, and moves pinned artifact IDs
 from `artifactAi` to `artifacts`. Saving stays disabled until the saved workspace
 has been successfully validated and restored. A loading failure shows an error
 and preserves the existing database snapshot for recovery.
+After successful restoration, migrations and startup changes retained by the
+merge are saved automatically if they differ from the loaded snapshot.
 
 Standalone chart blocks reuse the same Mosaic chart view and settings panel as
 dashboard charts. Charts with the same `selectionGroupId` in one Document share
--- a/apps/sqlrooms-cli-ui/README.md
+++ b/apps/sqlrooms-cli-ui/README.md
@@ -217,6 +217,12 @@ They can contain editable text, images, standalone Mosaic/vgplot chart blocks, a
 direct stateful blocks such as dashboards, pivot tables, Data Table Explorers,
 SQL queries, and Markdown documents.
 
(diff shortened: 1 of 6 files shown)


ilyabo:
Addressed the startup persistence feedback in 362798163.

`completeHydration` now compares the successfully restored state with the snapshot actually loaded from DuckDB and schedules a save when they differ. It no longer marks unwritten startup changes as saved. Completion reads the latest store state after AI/artifact synchronization and registers final-flush handlers, so those changes are included too. Failed restoration still leaves writes disabled; unchanged snapshots do not cause redundant writes.

Added regression coverage for startup changes surviving delayed loads of both new and existing workspaces, exactly one catch-up save, dirty-state bookkeeping, and post-load synchronization. All 18 targeted tests and the CLI typecheck pass. Added TSDoc for the persistence completion contract and migration helper, and updated the README.

PR #922, opened 2026-09-15
feat(kepler): expose per-map pending config status

ilyabo:
## Problem

A map with unavailable datasets can remain protected from autosave after restoration finishes. Hosts need a per-map status to warn users that their edits are not being saved.

## Changes

- Add `kepler.isMapConfigPending(mapId)`, derived from the existing pending-config guard without additional loading state.
- Document direct use in Zustand selectors and the distinction from the global persistence pause and overall readiness. Unknown, unregistered, and deleted maps return `false`.
- Test reactive status changes, independent maps, missing data and retry, and registration/deletion boundaries.

## Validation

- All 30 Kepler tests pass; all 34 repository test tasks pass.
- Kepler build, repository typechecks, Knip, and circular dependency checks pass.
- Changed-file ESLint passes with two existing warnings; formatting and `git diff --check` pass.

## Dependency

Follow-up to #920, which has merged. This PR targets `main` and contains only the new status API, documentation, and tests.

## Summary by CodeRabbit

- **New Features**
  - Added a reactive helper to check whether a saved map’s configuration or required data is still pending restoration.
  - The helper returns `false` for unknown or deleted maps and can be used with state selectors.

- **Documentation**
  - Documented the helper’s behavior, including restoration status and limitations.

- **Tests**
  - Added coverage for pending-state transitions during hydration, dataset restoration, map deletion, and persistence-paused restores.


xpectro93 (approved):
Thanks for turning this around so quickly. 30/30 kepler tests pass for me locally on this branch.

Deriving the status from `hasPendingKeplerConfig` rather than tracking it separately is the part worth calling out: it means the value a host reads is computed from the same guard the autosave middleware consults, so the two cannot drift. A cached flag would have been easier and would eventually have lied about whether a map was actually being saved.

Returning `false` for unknown and deleted maps is the right call too. It reads as "nothing of yours is going unsaved", which is what a host would act on, and the alternative — distinguishing "not pending" from "no such map" — would push a case onto every caller that none of them want.

On our side this replaces a diagnostic we had to hand-roll. We were watching the persisted config for a map's layer count dropping to zero and naming the recent kepler actions, purely because the failure was otherwise invisible. With #920 stopping the loss and this exposing the degraded state directly, that watchdog stops being worth keeping — we can query the store instead of inferring from a symptom, and surface something real to the user rather than a console warning. We have filed the removal on our side.

One logistical note rather than a review point: this targets `fix/kepler-config-hydration`, so it will want retargeting once #920 lands, as your description says. Worth not letting it sit too long after that merge.

PR #932, opened 2026-09-21
fix(ai-core): keep the default session name when title generation fails

xpectro93:
## Problem

`ai.sendPrompt` reports a failed model call by **resolving** with a placeholder string rather than throwing:

```ts
} catch (error) {
  console.error('Error generating text:', error);
  return AI_GENERATION_FAILED_TEXT; // 'error: can not generate response'
}
```

`generateSessionTitle` passed that response straight through to `renameSession`. The placeholder is non-empty and survives `cleanGeneratedSessionTitle`, so whenever the title call failed the chat was **renamed to `error: can not generate response`**. The hook's `try/catch` never fired, because nothing threw.

Worse, the session was then left with a non-default name, so `isDefaultSessionName` refused to auto-title it again — one transient model failure permanently stuck the chat with the error text.

## Fix

**Detect the failure out-of-band.** `sendPrompt` gains an optional `onError` callback, invoked before the placeholder is returned. `generateSessionTitle` branches on that rather than comparing the response text, so a conversation that legitimately produces the placeholder wording as its title is still renamed normally. The option is additive — no existing caller changes — and it composes with a host-supplied `onError` rather than replacing it.

**Fall back to `Untitled Chat`.** On failure the session is renamed to `Untitled Chat` instead of being left on its creation name, so it reads as untitled rather than as a brand-new chat.

Two supporting changes make that safe:

- `/^Untitled Chat$/` and `/^Untitled Chat \d+$/` join `DEFAULT_SESSION_NAME_PATTERNS`, so `isDefaultGeneratedSessionName` still matches and the **next user message can replace the fallback with a real title**. Without them the rename would quietly opt the chat out of auto-titling for good — the same stuck-name failure this PR is fixing, with friendlier wording.
- An optional `existingSessionNames`, supplied by the hook, numbers repeats (`Untitled Chat 1`, …) so two failures do not both land on the same name.

Verified end to end: `Chat` → failure → `Untitled Chat` → next message → real title.

## Notes on the approach

**Why not make `sendPrompt` throw?** That is the cleaner contract — a function that cannot signal failure except by value forces every caller to string-match — but it is a breaking change to a public API, so this PR takes the contained route.

**Why not a high-entropy sentinel?** The placeholder is user-visible: `ContextUsageIndicator` feeds `sendPrompt` output into `setPrompt`, so a marker like `__AI_GENERATION_FAILED_a8f3__` would surface in the composer. It stays human-readable, and `onError` carries the signal instead. The constant is internal to the package for the same reason — callers should use `onError`, not match on text.

**A caller's `onError` cannot break the contract.** It is invoked inside a `try/catch`, so a throwing callback is logged rather than turning the documented placeholder response into a rejection.

**`sessions` is read through a ref** in the hook. As a dependency of `generateTitle` it would change identity whenever any session changed, and the debounce effect depends on that callback, so unrelated session activity would clear the pending timer.

**One thing left alone:** `ContextUsageIndicator` also consumes `sendPrompt` output as content, so on failure it seeds a new chat with the placeholder text. Different surface, deserves its own change.

## Tests

`packages/ai-core/__tests__/generateSessionTitle.test.ts`, 7 cases:

- a failed generation renames to `Untitled Chat` and reports `generation-failed`
- a real title still renames as before
- a successful response that **equals** the placeholder is still applied as a title
- end-to-end against a real `createAiSlice` store whose model throws, driving the actual `sendPrompt`
- a caller `onError` that throws still yields the fallback rather than a rejection
- repeats are numbered `Untitled Chat 1`
- the fallback name remains eligible for a later retry

The end-to-end cases matter: a test that mocked `sendPrompt` with the placeholder would pass even if producer and check diverged. The throwing-callback case fails without its guard.

## Docs

`packages/ai-core/README.md` and `packages/ai/README.md` updated, per the repo's rule on public API changes.

## Verification

`packages/ai-core` tests pass with no regressions, `tsc --noEmit` introduces no new errors, prettier and eslint clean.

Not run locally: `knip`, `check-circular-deps`, and the full `turbo test` — this repo's `engines.node` requires `>=24.9` and the machine used was on Node 22, so the husky hooks could not execute. CI covers them and is green.


packages/ai-core/src/constants.ts
```diff
@@ -5,3 +5,12 @@ export const ABORT_EVENT = 'abort';
...
 export const ANALYSIS_CANCELLED = 'Analysis cancelled';
+
+/**
+ * What {@link AiSliceState.ai.sendPrompt} resolves with when the underlying
+ * model call fails. It reports failure by value rather than by throwing, so a
+ * caller that renders the response as content or as a name has to recognise
+ * this string — otherwise the placeholder is indistinguishable from a real
+ * answer. Compare against this constant rather than the literal.
+ */
+export const AI_GENERATION_FAILED_TEXT = 'error: can not generate response';
```
ilyabo, line 16: maybe we should just use 'Untitled Chat' for this situation

ilyabo (approved):

packages/ai-core/src/constants.ts
xpectro93: I looked into this and it has two side effects worth weighing before we adopt it.

**It would stop the chat ever being auto-titled again.** `isDefaultSessionName` gates retries on `DEFAULT_SESSION_NAME_PATTERNS`:

```
/^Chat$/  /^Chat \d+$/  /^Untitled$/  /^Default Session$/  /^Session /  /^\d{4}-\d{2}-\d{2} \d{2}:\d{2}$/
```

`'Untitled Chat'` matches none of them — `/^Untitled$/` requires exactly `Untitled`. So after one transient failure `generateSessionTitle` would return `custom-title` on every later message and never retry. That is the same stuck-name failure this PR is fixing, just with friendlier wording.

**The names would collide.** `renameSession` does a bare `session.name = name` with no dedup, unlike session creation which goes through `generateUniqueName('Chat', existingNames, ' ')`. Two failed chats would both read `Untitled Chat` in the sidebar.

Both are fixable — add `/^Untitled Chat( \d+)?$/` to the patterns and route the rename through `generateUniqueName`. The question is what it buys us: a new session is already named `Chat` / `Chat 2`, so on failure the user sees a sensible default either way. Not renaming keeps that name and leaves the retry alive on the next message.

Happy to switch to `Untitled Chat` with those two adjustments if you prefer the clearer signal — just wanted to flag that doing it without them would quietly disable the retry.

xpectro93: Went with your suggestion — the session is now renamed to Untitled Chat on failure rather than being left as Chat.

Two supporting changes make it safe:

- /^Untitled Chat$/ and /^Untitled Chat \d+$/ added to DEFAULT_SESSION_NAME_PATTERNS, so isDefaultGeneratedSessionName still matches and the next user message can replace it with a real title. Without this the rename would have permanently opted the chat out of auto-titling.

- An optional existingSessionNames on generateSessionTitle, passed by the hook, so repeats become Untitled Chat 1 instead of several identically named chats.
Verified end to end: Chat → failure → Untitled Chat → next message → real title.

